PR #77, opened 2022-08-10
AB Engine Flex integration

lyaunzbe:
Add Engine Flex event handlers for working with project external asset dependencies

This PR:
- creates or updates (if already existing) a `ProjectExternalAssetDependency` entity  when ABEngineFlex emits a `ExternalAssetDependencyUpdated` event
- removes a `ProjectExternalAssetDependency` entity when ABEngineFlex emits a `ExternalAssetDependencyRemoved` event. This works based on the assumption that the element being removed is the last one in the set (mapping), which the ABEngineFlex contract will always satisfy. We then proceed to refresh the remaining assets directly from the contract, avoiding any potential weirdness with multiple events in the same block.
- sets/updates preferredGateways (preferredIPFSGateway/preferredArweaveGateway) on the `Contract` entity when ABEngineFlex emits a `GatewayUpdated` event
- toggles externalAssetDependenciesLocked to true on the `Project` entity when the ABEngineFlex contract emits a `ProjectExternalAssetDependenciesLocked ` event. Note that this is a one way operation on the contract.
  
## Tests
- Tests are added for coverage of all new event handlers in pbab-mapping


lyaunzbe:
Ready for review, but DO NOT MERGE before config is updated with final contract deployment addresses. Will make comment with those addresses here once they are deployed and have been added

jakerockland (approved):

src/pbab-mapping.ts
```diff
@@ -176,6 +188,123 @@ export function handleTransfer(event: Transfer): void {
...
+
+  // Refresh project external asset dependencies
+  for (
+    let index = BigInt.fromI32(0);
+    index < lastEntityIndex;
+    index = index.plus(BigInt.fromI32(1))
+  ) {
+    const assetEntity = new ProjectExternalAssetDependency(
+      generateProjectExternalAssetDependencyId(project.id, index.toString())
+    );
```
ryley-o, line 257: I think all existing ProjectExternalAssetDependencies need to be removed from the store before creating the new ones here. This might appear to work on the surface, but I think there will be multiple entities at the same ID under the hood.
TBH, I'm fuzzy on how The Graph behaves when entity is created at the same ID, I think it gets weird, but maybe would just overwrite? Either way, if the last item was not the one removed, we would still have a stale item in the store, even if The Graph overwrites items. We had this type of issue when tx hash was our id for sales entities, and multiple token sales of a given token were happening in the same tx.

src/pbab-mapping.ts
```diff
@@ -176,6 +188,123 @@ export function handleTransfer(event: Transfer): void {
...
+  if (!contractEntity) {
+    return;
+  }
+  const dependencyType = FLEX_CONTRACT_EXTERNAL_ASSET_DEP_TYPES[event.params._dependencyType];
+  if (dependencyType === "IPFS") {
+    contractEntity.preferredIPFSGateway = event.params._gatewayAddress;
+  } else {
+    contractEntity.preferredArweaveGateway = event.params._gatewayAddress;
+  }
+  contractEntity.save();
```
ryley-o, line 291: update `updatedAt`

tests/subgraph/pbab-mapping/pbab-mapping.test.ts
```diff
@@ -1842,6 +1862,306 @@ test("GenArt721Core2PBAB: Can handle transfer", () => {
...
+  assert.fieldEquals(
+    PROJECT_ENTITY_TYPE,
+    fullProjectId,
+    "externalAssetDependencyCount",
+    _externalAssetDependencyCount1.toString()
+  );
+  
+});
+
+test("GenArt721Core2EngineFlex: Can remove a project external asset dependency", () => {
```
ryley-o, line 1983: I think adding another test that confirms e.g. removing external asset at index 0 after adding two external assets, then using `assert.entityCount` to confirm there are no duplicate entities in store, would be 🔥

ryley-o (commented):
Woohoo! A couple comments on removing stale items from store before refreshing. 🙏

src/pbab-mapping.ts
lyaunzbe: good call on this 👍

schema.graphql
```diff
@@ -414,16 +426,39 @@ type SaleLookupTable @entity {
...
+
+  "The dependency type"
+  dependencyType: ProjectExternalAssetDependencyType!
+
+  "The dependency cid"
+  cid: String!
+
+  "The dependency index"
+  index: BigInt!
+}
```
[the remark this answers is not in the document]
lyaunzbe: ya I think it couldn't hurt (to distinguish from updatedAt at the project level, which may be because of an external assets dep operation but may not be)

tests/subgraph/pbab-mapping/pbab-mapping.test.ts
lyaunzbe: sounds good, but couldn't that assertion be preformed with this test as well/not sure why we would need a new separate test?

src/pbab-mapping.ts
```diff
@@ -176,6 +188,123 @@ export function handleTransfer(event: Transfer): void {
...
+
+export function handleExternalAssetDependencyUpdated(
+  event: ExternalAssetDependencyUpdated
+): void {
+  let project = Project.load(
+    generateContractSpecificId(event.address, event.params._projectId)
+  );
+
+  if (!project) {
+    return;
```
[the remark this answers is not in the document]
lyaunzbe: i dont think this is being done with any other handlers in this mapping, but personally this seems worth it to me/not that much clutter for the benefit

schema.graphql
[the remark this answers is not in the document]
lyaunzbe: no strong opinion on this, but thinking more on it, `updatedAt` would seem most useful for me tracking against all externalAssetDep entities, vs each individual one.  Are there any use-cases you can think of of that perhaps I'm missing?

tests/subgraph/pbab-mapping/pbab-mapping.test.ts
lyaunzbe: EDIT: I actually got confused here with the contract tests, where I already do explicitly test removal after multiple adds. This test is currently only adding one and removing it. I still think we can use a single test here, will just update this current one to add more than one dep!

src/pbab-mapping.ts
lyaunzbe: Can confirm, from testing, that the graph overwrites entities as one would expect. Tests also confirm that multiple entities at the same ID do not get left behind.

> Either way, if the last item was not the one removed, we would still have a stale item in the store, even if The Graph overwrites items. We had this type of issue when tx hash was our id for sales entities, and multiple token sales of a given token were happening in the same tx.

Because of the way the removal operation is written in the smart contract, it's ALWAYS the last item removed. I've added a comment above the handler to make sure this is very clear. I've also added to the removal event handler tests to make sure that there are no extraneous entities that get left and that the shifted data is where it should be

tests/subgraph/pbab-mapping/pbab-mapping.test.ts
lyaunzbe: added 👍

schema.graphql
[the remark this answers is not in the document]
ryley-o: No strong opinions from me here 👍 just didn't want to forget it if you wanted it 👍

src/pbab-mapping.ts
ryley-o: ah, yes, thanks for helping me understand the overwrite behavior here! 🙏 

My original comment about "we would have a stale item regardless of overwrite behavior" is definitely incorrect - a "stale" item would only exist if same ID didn't overwrite. Thus, this LGTM and appreciate the extra test check to confirm this behavior 🔥

ryley-o (approved):

src/pbab-mapping.ts
```diff
@@ -9,6 +9,7 @@ import {
   ByteArray
 } from "@graphprotocol/graph-ts";
 
+import { log as logger } from "matchstick-as";
```
yoshiwarab, line 12: Nit: remove please

tests/subgraph/pbab-mapping/pbab-mapping.test.ts
```diff
@@ -1842,6 +1863,358 @@ test("GenArt721Core2PBAB: Can handle transfer", () => {
...
+
+  // checks project external asset dependency lock status
+  assert.fieldEquals(
+    PROJECT_ENTITY_TYPE,
+    fullProjectId,
+    "externalAssetDependenciesLocked",
+    "true"
+  );
+});
+
```
yoshiwarab, line 2216: Can you add a quick test for the case where no project exists for all the handler?

yoshiwarab (approved):

PR #88, opened 2022-09-01
V3/Handle propose/accept artist addresses and splits

ryley-o:
Add V3 core event handlers for proposing and accepting new artist addresses and splits.

This PR:
- Creates a new `ProposedArtistAddressesAndSplits` entity when V3 core emits `ProposedArtistAddressesAndSplits` event
  - also removes any existing `ProposedArtistAddressesAndSplits` entity at the same ID, since this process works by overwriting any existing proposals on the blockchain
- Sets project addresses and splits to current proposed values when V3 core emits `AcceptedArtistAddressesAndSplits`
  - also clears existing proposed addresses and splits, aligning with the logic in this PR: https://github.com/ArtBlocks/artblocks-contracts/pull/279
- Adds a `createdAt` field to the `ProposedArtistAddressesAndSplits` entity in our schema
  - just something I forgot to add previously
  
## Tests
- Tests are added for coverage of new handlers

### DIFF
--- a/src/mapping-v3-core.ts
+++ b/src/mapping-v3-core.ts
@@ -22,7 +22,8 @@ import {
   AccountProject,
   ProjectScript,
   Contract,
-  MinterFilter
+  MinterFilter,
+  ProposedArtistAddressesAndSplits
 } from "../generated/schema";
 
 import {
--- a/src/mapping-v3-core.ts
+++ b/src/mapping-v3-core.ts
@@ -15,7 +15,9 @@ import {
   Mint,
   Transfer,
   PlatformUpdated,
-  MinterUpdated
+  MinterUpdated,
+  ProposedArtistAddressesAndSplits as ProposedArtistAddressesAndSplitsEvent,
+  AcceptedArtistAddressesAndSplits
 } from "../generated/GenArt721CoreV3/GenArt721CoreV3";
 
 import { MinterFilterV0 } from "../generated/MinterFilterV0/MinterFilterV0";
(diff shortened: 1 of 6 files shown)


src/mapping-v3-core.ts
```diff
@@ -552,6 +555,110 @@ export function handleMinterUpdated(event: MinterUpdated): void {
   contractEntity.save();
 }
 
+// Handle artist proposed address and splits updates
+// This is an event indicating that the artist has proposed a new set of
+// addresses and splits for a project.
+export function handleProposedArtistAddressesAndSplits(
```
yoshiwarab, line 561: Totally fine to leave the code as is but we could reduce the lines of code here by not clearing the store between each test. If you decide to go this route I'd suggest wrapping everything here in a describe block to give it some explicit logical grouping.

Also, can you add a couple tests for the early return cases.

yoshiwarab (approved):

src/mapping-v3-core.ts
```diff
@@ -240,6 +242,110 @@ export function handleMinterUpdated(event: MinterUpdated): void {
...
+  const project = Project.load(newEntityId);
+  if (!project) {
+    return;
+  }
+  // remove any existing proposed artist addresses and splits
+  const existingProposedArtistAddressesAndSplitsId =
+    project.proposedArtistAddressesAndSplits;
+  if (existingProposedArtistAddressesAndSplitsId !== null) {
+    project.proposedArtistAddressesAndSplits = null;
+    store.remove(
```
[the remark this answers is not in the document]
ryley-o: @yoshiwarab actually, given my updated understanding of how overwriting works, do you think it would be okay to remove this entire `//remove any existing proposed artist addresses and splits` code block? When we create a new entity, it should just automatically overwrite any existing entity.

src/mapping-v3-core.ts
yoshiwarab: @ryley-o this was meant to be in the test file oops!

src/mapping-v3-core.ts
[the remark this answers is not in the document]
yoshiwarab: Ah yeah that works even better! As long as all of the fields are overwritten (if any are not they'll keep the old values)

src/mapping-v3-core.ts
ryley-o: got it 👍 good call, updated with describe block and a few tests for the early return cases.

src/mapping-v3-core.ts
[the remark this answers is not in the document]
ryley-o: SGTM! good call about all fields needing to be updated - we are good here because all fields will always be populated from the event. updated and pushed!

src/mapping-v3-core.ts
ryley-o: (just to note, I kept the`clearStore` in a BeforeEach, just to help me keep things clean between the different tests)

lyaunzbe (approved):

PR #93, opened 2022-09-07
Bug fix: JSON Bytes string validity check

ryley-o:
Fix bug in `bytesToJSONValue` generic handler logic used to determine if bytes value is an encoded UTF-8 string vs. raw bytes.

A false-positive "UTF-8 encoded string" was detected in our test suite in some unrelated work. Specifically, the randomly generated address `0x57e32bd396b7c3337dd6b4a672e6d99b47865e56` was tested and found to be a valid "WTF-8 encoded string", which is used internally by AssemblyScript, but not a valid "UTF-8 encoded string". Upon further investigation, the likelihood of of us attempting to upload a bytes32 not intended to be an encoded string (e.g. Merkle root) is pretty high (Based on testing a few random addresses, I would estimate anywhere from 10% to 50%, not completely sure without spending more time).

This updates our handler logic to convert from `Bytes` to string via `Bytes.toString()` method, then from string to Bytes via `string.toHexString()` method. The starting `Bytes` and the decoded/re-encoded `Bytes` will only be the same if the original `Bytes` was a valid `UTF-8` encoded string.

Note that it is possible for this to detect false-positives and improperly assume bytes are an encoded string. The likelihood of that occurring goes up with shorter bytes values being stored. When using a Bytes32, it seems negligibly likely that we would encounter a false positive, If we want to prove that we should do some more work to quantify.

## Deployment Impacts
This bug fix should be included for any work with the new Merkle minter (currently in Goerli-dev deployment) Cc @mchrupcala

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -177,10 +177,21 @@ export function arrayToJSONValue(value: string): JSONValue {
 // If byte data is parseable to a valid unicode string then do so
 // otherwise parse the byte data to a hex string
 export function bytesToJSONValue(value: Bytes): JSONValue {
-  // If the bytes cannot be
-  let result = json.try_fromString('["' + value.toString() + '"]');
-  if (result.isError) {
-    result = json.try_fromString('["' + value.toHexString() + '"]');
+  // fallback - assume the data is a hex string (always valid)
+  let result = json.try_fromString('["' + value.toHexString() + '"]');
+  // If the bytes can be parsed as a string, then losslessly re-encoded into
+  // UTF-8 bytes, then consider a valid UTF-8 encoded string and store
+  // string value in json.
+  // note: Bytes.toString() uses WTF-8 encoding as opposed to UTF-8.  Solidity
+  // encodes UTF-8 strings, so safe assume any string data are UTF-8 encoded.
+  let stringValue: string = value.toString();
+  let reEncodedBytesValue: Bytes = Bytes.fromUTF8(stringValue);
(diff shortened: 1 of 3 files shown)


ryley-o:
Also Cc @yoshiwarab for visibility 👍

ryley-o:
@Jbern16 just realized - does this need to also be considered in `handleAddManyValueGeneric` in `minter-suite-mapping.ts`?

Jbern16 (approved):

Jbern16:
I think we should be good since we are using that method here - https://github.com/ArtBlocks/artblocks-subgraph/blob/main/src/minter-suite-mapping.ts#L556

Jbern16:
Oh wait you are right - its the many handler that's not explicit about this it seems - https://github.com/ArtBlocks/artblocks-subgraph/blob/main/src/minter-suite-mapping.ts#L724-L731. That probably should updated to use this method

ryley-o:
@Jbern16 cool cool I'll update the many handler to use this, and ping you when updated 👍

jakerockland (approved):

ryley-o:
@Jbern16 updated!
This has sufficient approvals, so I'm going to merge 👍

ryley-o:
@lyaunzbe not a big rush, but this will only merge after your approval 👍

lyaunzbe (approved):

lyaunzbe:
appreciate the detailed writeup + comments here 💯

PR #96, opened 2022-09-12
Emulate Set instead of Array for generic "many values"

ryley-o:
Update `handleAddManyValueGeneric` handler to treat "many values" as a Set instead of an Array.

I _think_ this is desired behavior, but would love input from others on this.

Current behavior:
- Behave like Array:
  - e.g. If `100` is added twice, generic json would be `{<key>: [100, 100]}`
Updated behavior:
- Behave like Set:
  - e.g. if `100` is added twice, generic json would be: `{<key>: [100]}`

A Set is typically preferable because how Solidity works so well with mappings, which act more like Sets than Arrays. Emulating a Set in our subgraph eliminates on-chain validation to protect against adding something more than once (reducing gas cost and on-chain complexity). Most relevant current example is [this line in our MinterHolder](https://github.com/ArtBlocks/artblocks-contracts/blob/ed09f1c352916d0736b06ac7eb913b157aef7b51/contracts/minter-suite/Minters/MinterHolder/MinterHolderV0.sol#L197)
With this update, smart contract doesn't have to validate if the project is already allowlisted, it simply sets the mapping value to true, and emits an event indicating the project is allowlisted. It is up to the frontend to intelligently avoid duplicate allow/deny requests when queueing up txs.

Cc @mchrupcala since this was based on convo from today 🙏

### DIFF
--- a/src/minter-suite-mapping.ts
+++ b/src/minter-suite-mapping.ts
@@ -707,7 +707,10 @@ export function handleAddManyValueGeneric<T, C>(
     if (val) {
       arr = val.toArray().map<BigInt>((v: JSONValue) => v.toBigInt());
     }
-    arr.push(event.params._value);
+    // only add if it doesn't exist, so we act like a Set
+    if (!arr.includes(event.params._value)) {
+      arr.push(event.params._value);
+    }
     newValue = arrayToJSONValue(arr.toString());
   } else if (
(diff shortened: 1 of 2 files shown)


jakerockland (approved):
This tracks for me.

mchrupcala (approved):

lyaunzbe (approved):

PR #100, opened 2022-09-15
Index AdminACLV0 superAdmin as V3 core's owner and whitelisted

ryley-o:
Add indexing of AdminACLV0 contracts to forward their `superAdmin` address to the V3 core's `admin` and `whitelisted` fields.

Before: contract entity's admin was set to the adminACL address
After: contract entity's admin and whitelisted is set to the single `superAdmin` address from the contract's AdminACLV0 reference. If ownership is renounced, the contract entity's admin is set to zero address, and whitelisted is empty array.

Note that this relies on the `superAdmin` properly notifying which contracts should be refreshed when the `changeSuperAdmin` is called on the AdminACLV0 contract, as described in this PR: https://github.com/ArtBlocks/artblocks-contracts/pull/289. Also note that worst-case if a core isn't included in the `changeSuperAdmin` call, temporarily the subgraph isn't aware of the new superAdmin, and any event that triggers `refreshContract` will eventually refresh the contract to the appropriate admin & whitelisted.

This corresponds to the changes in https://github.com/ArtBlocks/artblocks-contracts/pull/289.

Tests are added for the new functionality.


ryley-o:
Cc @Jbern16

ryley-o:
gentle ping to @lyaunzbe @yoshiwarab on this one - once this is merged, I should be able to get a subgraph deployment for V3 out that works for @Jbern16 and myself (frontend + sync server work).

src/helpers.ts
```diff
@@ -48,6 +53,21 @@ export function generateContractSpecificId(
...
+  account.save();
+
+  let whitelisting = new Whitelisting(
+    generateWhitelistingId(contractId, account.id)
+  );
+  whitelisting.account = account.id;
+  whitelisting.contract = contractId;
+
+  whitelisting.save();
+  return whitelisting.id;
```
yoshiwarab, line 68: Any reason to not just return the full whitelisting entity here?

subgraph.template.yaml
```diff
@@ -1243,6 +1243,31 @@ dataSources:
...
+      abi: AdminACLV0
+      startBlock: {{startBlock}}
+    mapping:
+      kind: ethereum/events
+      apiVersion: 0.0.5
+      language: wasm/assemblyscript
+      entities:
+        - Project
+        - Contract
+      abis:
```
yoshiwarab, line 1261: I think you'll need to include the GenArt721CoreV3 in the ABIs list here as well since you make GenArt721CoreV3 contract calls when the SuperAdminTransferred event handler gets called. Don't think this breaks the tests but will break when actually running.

tests/subgraph/mapping-v3-core/mapping-v3-core.test.ts
```diff
@@ -825,6 +911,62 @@ test("GenArt721CoreV3: Handles PlatformUpdated::artblocksSecondaryBPS - changed
   );
 });
 
+describe("GenARt721CoreV3: handleIAdminACLV0SuperAdminTransferred", () => {
```
yoshiwarab, line 914:
```suggestion
describe("GenArt721CoreV3: handleIAdminACLV0SuperAdminTransferred", () => {
```

yoshiwarab (approved):

subgraph.template.yaml
ryley-o: SGTM! updated and pushed! I also added the IAdminACLV0 to the core contract portion of the template, but maybe we can chat more about exactly why these are needed (I don't have a great mental model of what these ABIs do...)

src/helpers.ts
ryley-o: good call! no reason - updated to return entity instead of entity id 👍

lyaunzbe (approved):

PR #106, opened 2022-09-22
Flex related bugfix + dev goerli config and flex ABI updated

lyaunzbe:
minor bugfix related to Flex contract after local testing:
- casting externalDependencyAssetCount to BigInt

also, after merging https://github.com/ArtBlocks/artblocks-contracts/pull/303 we have updated Flex contract ABI + re-generated codegen + updated dev goerli with new deployment addresses


src/mapping-v2-core.ts
```diff
@@ -288,12 +295,13 @@ export function handleExternalAssetDependencyRemoved(
...
   if (!contractEntity) {
     log.warning("Contract not found for GatewayUpdated event", []);
     return;
   }
-  const dependencyType = FLEX_CONTRACT_EXTERNAL_ASSET_DEP_TYPES[event.params._dependencyType];
+  const dependencyType =
+    FLEX_CONTRACT_EXTERNAL_ASSET_DEP_TYPES[event.params._dependencyType];
   if (dependencyType === "IPFS") {
     contractEntity.preferredIPFSGateway = event.params._gatewayAddress;
   } else {
```
ryley-o, line 307: Do we need to clear the dependency not being set (e.g. clear arweave if setting IPFS)?

ryley-o (approved):
🙌 🙌  LGTM

src/mapping-v2-core.ts
lyaunzbe: a project can have assets from both IPFS and Arweave, so I don't think we'd want to clear

ryley-o: ahhhh makes sense! SGTM!

yoshiwarab (approved):

PR #109, opened 2022-09-26
Remove curationStatus/curation_status handling from subgraph entirely

jakerockland:
A better solution to on-chain curation status is being pursued via curation registry efforts.

For now, this data should be most reliably pulled from Art Blocks APIs (the Token API currently, and additionally the GraphQL API when released in the near future).

### DIFF
--- a/generated/schema.ts
+++ b/generated/schema.ts
@@ -229,23 +229,6 @@ export class Project extends Entity {
     }
   }
 
-  get curationStatus(): string | null {
-    let value = this.get("curationStatus");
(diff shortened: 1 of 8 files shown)


jakerockland:
With https://github.com/ArtBlocks/artblocks/pull/1075 out, I think this is ready for review @ryley-o @yoshiwarab @lyaunzbe (just needs to wait to be merged/deployed until after sync server update is shipped).

lyaunzbe (approved):

PR #120, opened 2022-10-06
Add details for initial Flex deployments (Bright Moments - Flex; Verse)

jakerockland:
Hasn't been deployed yet, as @ryley-o was already planning to do an artist-staging deployment so I'm going to let him tag this along there, but this adds the subgraph details for:

- https://github.com/ArtBlocks/artblocks/issues/287
- https://github.com/ArtBlocks/artblocks/issues/275


config/goerli-staging.json
```diff
@@ -53,6 +53,14 @@
       "address": "0x3b30d421a6dA95694EaaE09971424F15Eb375269",
       "name": "Endaoment Gallery for Good"
     },
+    {
+      "address": "0x2146a41f2c1432895d6d0adf9c60bf0a226bac0e",
+      "name": "Bright Moments - Flex"
+    },
+    {
+      "address": "0xa319c382a702682129fcbf55d514e61a16f97f9c",
+      "name": "Verse"
```
ryley-o, line 62: nit: can we call this Verse - Flex?

ryley-o (approved):

config/goerli-staging.json
jakerockland: I was making this match the actual names of the contracts as configured at deployment, which is why BMG includes it (their preference) and Verse doesn't.

That said, can definitely use our own naming conventions here.. but wanted to flag this in case the additional context changed your opinion here on what makes sense.

WDYT?

ryley-o: oh yeah, this is in the flex bucket already - resolving!

lyaunzbe (approved):

PR #135, opened 2022-11-18
Dev Deployment for MinterMerkleV3 & MinterHolderV2

mchrupcala:
## Description of the change

This PR integrates the MerkleV3 & HolderV2 contracts (both integrating the logic for [vault delegation](https://delegate.cash/)) to dev (only).

We also have 2 new event handlers for when the delegate.cash registry address is set (this only happens in the constructor during contract deployment) -- we'll store this address in extraMinterDetails at the minter-level. Relevant handlers are `handleMerkleDelegationRegistryUpdated` and `handleHolderDelegationRegistryUpdated`.


src/minter-suite-mapping.ts
```diff
@@ -570,17 +594,66 @@ export function handleRegistrationNFTAddresses<T>(event: T): void {
...
 
+export function handleMerkleV3DelegationRegistryUpdated(
+  event: MinterMerkleV3DelegationRegistryUpdated
+): void {
+  let genericEvent: ConfigValueSetAddress;
+  genericEvent = changetype<ConfigValueSetAddress>(event);
+
+  genericEvent.parameters = [
+    // dummy value ("project ID: 0") because we're not passing in a project
+    new ethereum.EventParam("_projectId", ethereum.Value.fromI32(0)),
```
mchrupcala, line 618: Since we're using `handleSetValueGeneric` at the minter level (and since there's no project), we had to pass in a dummy projectID here. Just wanted to call that out in case ya'll had any concerns about it
CC: @ryley-o

subgraph.template.yaml
```diff
@@ -1311,6 +1355,44 @@ dataSources:
...
+        - event: ProjectCurrencyInfoUpdated(indexed uint256,indexed address,string)
+          handler: handleProjectCurrencyInfoUpdated
+        - event: ConfigValueSet(indexed uint256,bytes32,bytes32)
+          handler: handleSetBytesValueProjectConfig
+        - event: ConfigValueSet(indexed uint256,bytes32,bool)
+          handler: handleSetBooleanValueProjectConfig
+        - event: ConfigValueSet(indexed uint256,bytes32,uint256)
+          handler: handleSetBigIntValueProjectConfig
+        - event: DelegationRegistryUpdated(address)
+          handler: handleMerkleV3DelegationRegistryUpdated
```
ryley-o, line 1393: I just realized that this event actually comes from the `IFilteredMinterMerkleV1` interface.
Can you copy that ABI over to the abis directory, and import that event from the interface in `./src/minter-suite-mapping.ts` (and associated test files)?
(Basically, follow a similar pattern as `IFilteredMinterV0` and `PricePerTokenInWeiUpdated`)
This will prevent us from having to add a new handler if we ever release a new version of the Merkle minter.

subgraph.template.yaml
```diff
@@ -1239,6 +1243,46 @@ dataSources:
...
+        - event: RegisteredNFTAddress(indexed address)
+          handler: handleRegisteredNFTAddressV2
+        - event: UnregisteredNFTAddress(indexed address)
+          handler: handleUnregisteredNFTAddressV2
+        - event: AllowedHoldersOfProjects(indexed uint256,address[],uint256[])
+          handler: handleAllowHoldersOfProjectsV2
+        - event: RemovedHoldersOfProjects(indexed uint256,address[],uint256[])
+          handler: handleRemoveHoldersOfProjectsV2
+        - event: DelegationRegistryUpdated(address)
+          handler: handleHolderV2DelegationRegistryUpdated
```
ryley-o, line 1283: same comment as MinterMerkleV3 - can we make this come from the interface `IFilteredMinterHolderV1`?

ryley-o (approved):
TY @mchrupcala! Great work 🙏 
I noticed that we can use the interface for the new delegation registry events, so would be great to do that. If time is limited, could always do that as follow-on.

yoshiwarab (changes requested):
Looks great! Just a couple of optimizations

src/minter-suite-mapping.ts
```diff
@@ -95,6 +95,14 @@ import {
...
   RemovedHoldersOfProjects as MinterHolderV1RemovedHoldersOfProjects
 } from "../generated/MinterHolderV1/MinterHolderV1";
+import {
+  AllowedHoldersOfProjects as MinterHolderV2AllowedHoldersOfProjects,
+  RegisteredNFTAddress as MinterHolderV2RegisteredNFTAddress,
+  UnregisteredNFTAddress as MinterHolderV2UnregisteredNFTAddress,
+  RemovedHoldersOfProjects as MinterHolderV2RemovedHoldersOfProjects
+} from "../generated/MinterHolderV2/MinterHolderV2";
+import { DelegationRegistryUpdated as MinterHolderV2DelegationRegistryUpdated } from "../generated/MinterHolderV2/IFilteredMinterHolderV1";
+import { DelegationRegistryUpdated as MinterMerkleV3DelegationRegistryUpdated } from "../generated/MinterMerkleV3/IFilteredMinterMerkleV1";
```
yoshiwarab, line 105: Once redeployed, we should double check that this doesn't break. I think since the event is from the interface to begin with it should be fine but since the handlers are for the specific types of contracts I'm worried it might bork.

src/minter-suite-mapping.ts
mchrupcala: SGTM! I ran with this and went a little further, changing `handleSetValueGeneric` into `handleSetMinterDetailsGeneric`since since the JSON we're updating only lives on the minter. I also moved the [type-checking](https://github.com/ArtBlocks/artblocks-subgraph/pull/135/commits/a697ac6670cda6d5e3a82f864de8064d44157acf#diff-a1077c7a5a21bf6e3a06b3124ce379bd5374fef2987d7164b250ae14fad15c56R611) to a higher scope so that we'll only update `minter.updatedAt` if the key-value's been successfully saved.

src/minter-suite-mapping.ts
ryley-o: ah, yeah, should probably import from the interface here (I think)...

ryley-o: I think this can be resolved, with the agreement that this should be tested on our dev subgraph to ensure nothing breaks. The best way to test this is to ensure we either fully re-sync our dev subgraph without grafting, or ensure if we graft, the starting block is before these two minters were deployed (since they only emit this event during deployment).

ryley-o, line 105: nit: Since these events come from the minter interfaces instead of the specific version of Merkle/Holder minters, can we omit the version numbers, and instead import the Holder and Merkle events as `MinterHolderDelegationRegistryUpdated` and `MinterMerkleDelegationRegistryUpdated`, respectively?

src/minter-suite-mapping.ts
```diff
@@ -607,51 +607,39 @@ export function handleRegistrationNFTAddresses<T>(event: T): void {
...
+
+  handleSetMinterDetailsGeneric(
+    "delegationRegistryAddress",
+    event.params.delegationRegistryAddress,
+    minter
+  );
+  minter.updatedAt = event.block.timestamp;
+}
+
+export function handleMerkleV3DelegationRegistryUpdated(
```
ryley-o, line 634: nit: Since this is handling event from interface, should be able to remove the V3 from this function name

src/minter-suite-mapping.ts
```diff
@@ -607,51 +607,39 @@ export function handleRegistrationNFTAddresses<T>(event: T): void {
...
+  event: MinterMerkleV3DelegationRegistryUpdated
+): void {
+  handleDelegationRegistryUpdatedGeneric(event);
 }
 export function handleHolderV2DelegationRegistryUpdated(
```
ryley-o, line 639: nit: Since this is handling event from the interface, should be able to remove V2 from this function name

src/minter-suite-mapping.ts
```diff
@@ -692,54 +680,30 @@ export function handleUnregisteredNFTAddressV2(
...
+  if (isBoolean(value)) {
+    jsonValue = json.fromString(booleanToString(value));
+  } else if (value instanceof BigInt) {
+    jsonValue = json.fromString(value.toString());
+  } else if (value instanceof Address) {
+    jsonValue = stringToJSONValue(value.toHexString());
+  } else if (value instanceof Bytes) {
+    jsonValue = bytesToJSONValue(value);
+  } else {
+    return;
```
ryley-o, line 701: nit: we should log a warning here, since we don't expect to ever reach this point.

src/minter-suite-mapping.ts
```diff
@@ -692,54 +680,30 @@ export function handleUnregisteredNFTAddressV2(
 // Because AssemblyScript does not support union types, we need to manually type check inside each method, to ensure correct usage.
 // For any questions reach out to @jon or @ryley-o.eth. or see the following document https://docs.google.com/document/d/1XSxl04eJyTxc_rbj6cmq-j00zaYDzApBBLT67JXtaOw/edit?disco=AAAAZa8xp-Q
```
ryley-o, line 681: nit: can we update this function's documentation a bit? would be great to define what type of values are allowed for T, and what types of values are allowed for C.

src/minter-suite-mapping.ts
```diff
@@ -692,54 +680,30 @@ export function handleUnregisteredNFTAddressV2(
...
-  }
-
-  if (project) {
-    project.updatedAt = event.block.timestamp;
-  }
-
   let minterDetails = getMinterDetails(config);
-  let jsonKey = event.params._key.toString();
+  let jsonKey = key;
   let jsonValue: JSONValue;
```
ryley-o, line 691: should we add something like the following? (we might get errors when running the subgraph locally without?)
```
if (
    !(config instanceOf Minter ||
    config instanceOf Project)
) {
    log.warning("[WARN] Generic property attempted to be set on something not a Minter or Project")
    return;
}
```
Also, we might want to do something similar for value `<T>` (for the same reasons?)

src/minter-suite-mapping.ts
```diff
@@ -692,54 +680,30 @@ export function handleUnregisteredNFTAddressV2(
...
+    jsonValue = bytesToJSONValue(value);
+  } else {
+    return;
   }
 
   minterDetails.set(jsonKey, jsonValue);
-
   config.extraMinterDetails = typedMapToJSONString(minterDetails);
-
   config.save();
```
ryley-o, line 706: we should update the `updatedAt` field before saving here

src/minter-suite-mapping.ts
```diff
@@ -607,51 +607,39 @@ export function handleRegistrationNFTAddresses<T>(event: T): void {
...
+  if (!minter) {
+    return;
+  }
+
+  handleSetMinterDetailsGeneric(
+    "delegationRegistryAddress",
+    event.params.delegationRegistryAddress,
+    minter
+  );
+  minter.updatedAt = event.block.timestamp;
```
ryley-o, line 631: I think we should update this to populate the `updatedAt` field in `handleSetMinterDetailsGeneric` so we should remove this line here. (it isn't doing anything since we aren't saving here anyway)

src/minter-suite-mapping.ts
```diff
@@ -759,11 +723,24 @@ export function handleSetValueProjectConfig<T>(event: T): void {
...
+
+    handleSetMinterDetailsGeneric(
+      event.params._key.toString(),
+      event.params._value,
+      config
     );
+
+    if (minterProjectAndConfig.project) {
+      minterProjectAndConfig.project.updatedAt = event.block.timestamp;
+    }
```
ryley-o, line 743: This isn't doing anything since we don't save the project after this. However, I don't think we want to update the updatedAt on the project entity in this case - we just want the handleSetMinterDetailsGeneric to update the ProjectMinterConfiguration's updatedAt timestamp. So I think this can be removed?

tests/subgraph/minter-suite/minter-suite-mapping.test.ts
```diff
@@ -86,19 +86,33 @@ import {
...
   UnregisteredNFTAddress as HolderV1UnregisteredNFTAddress
 } from "../../../generated/MinterHolderV1/MinterHolderV1";
+import {
+  AllowedHoldersOfProjects as HolderV2AllowedHoldersOfProjects,
+  RemovedHoldersOfProjects as HolderV2RemovedHoldersOfProjects,
+  RegisteredNFTAddress as HolderV2RegisteredNFTAddress,
+  UnregisteredNFTAddress as HolderV2UnregisteredNFTAddress
+} from "../../../generated/MinterHolderV2/MinterHolderV2";
+import { DelegationRegistryUpdated as MinterHolderV2DelegationRegistryUpdated } from "../../../generated/MinterHolderV2/IFilteredMinterHolderV1";
+import { DelegationRegistryUpdated as MinterMerkleV3DelegationRegistryUpdated } from "../../../generated/MinterMerkleV3/IFilteredMinterMerkleV1";
```
ryley-o, line 96: nit: I think we can remove the V2 and V3 from these import-as names, since we are importing from the interfaces.

ryley-o (approved):
🙏 Thank you for the latest refactor, LGTM!
A few minor nits is all 🥳

src/minter-suite-mapping.ts
yoshiwarab: ++ Since we're updating the entity here, I do think the type check logic should also be here. @ryley-o `T` is type narrowed below so we don't need to do the extra type check but should definitely still warn like suggested.

Overall, adding Ryley's logic is enough here but thinking on this more, we should do some additional refactoring as a follow-up that should simplify/clarify the code a bit! Currently, this function is sort of halfway between a utility function and an actual handler, which makes it a little hard to understand where and what to update. Instead, we can break out setJSONValue that takes `key`, `value`, and `map` and returns the updated map.  I think we'd want to do something similar for all of the handlers that are mostly doing `TypedMap<string, JSONValue>` manipulation so that we can use extra_minter_details for custom events without having to manipulate them to fit the generic event format.

src/minter-suite-mapping.ts
yoshiwarab: This requires that the entity (config) is a `minter` and not a `projectMinterconfiguration` so we'd have to check that config is a minter first.

src/minter-suite-mapping.ts
yoshiwarab: Currently no updatedAt on the ProjectMinterConfiguration entity so we do actually want to set the updatedAt on the project here

yoshiwarab (approved):

src/minter-suite-mapping.ts
mchrupcala: Sure! I [moved this type-checking](https://github.com/ArtBlocks/artblocks-subgraph/pull/135/commits/a697ac6670cda6d5e3a82f864de8064d44157acf#diff-a1077c7a5a21bf6e3a06b3124ce379bd5374fef2987d7164b250ae14fad15c56R729) outside this function (to our handler) because without it, the `project.updatedAt` call would run even if `handleSetMinterDetailsGeneric` bombed due to a failed type-check. 

Makes sense that we'd want this check to live inside the Generic handler so I added Ryley's logic + I'll add a ticket to refactor this too :)

mchrupcala: Asana ticket for follow-up refactor
https://app.asana.com/0/1202106026831420/1203450133533228

src/minter-suite-mapping.ts
mchrupcala: For sure! I listed out the T values but it looks like we already called out the supported C values. But actually I noticed the documentation showed ProjectMinterConfiguration as a valid type -- since we're type-checking for Minter & Project [I changed this to "Project"](https://github.com/ArtBlocks/artblocks-subgraph/pull/135/commits/ccfdbd4ea1fdfd97e2a2c0b32961f15658670f28#diff-a1077c7a5a21bf6e3a06b3124ce379bd5374fef2987d7164b250ae14fad15c56R680) but I wanted to flag this in case we're missing one maybe?
@ryley-o

mchrupcala: Nvm I [changed this back](https://github.com/ArtBlocks/artblocks-subgraph/pull/135/commits/18a42c918ab3e9dc980f3a162fbe1f06919b4367) since it looks like we're only using this handler on Minters & ProjectMinterConfigs

PR #138, opened 2022-11-23
DA-w-Settlement/Add to subgraph template

ryley-o:
## Description of the change

This adds the DA w/Settlement minter to the subgraph template.

Associated details such as adding ABIs, placeholder to the dev config, and generated files are also included in this PR.

Event handlers specific to the new minter will be written in follow-on PR(s). Tests will be added in follow-on PR(s).


ryley-o:
as discussed, smol update to reflect settlement nomenclature instead vs. refund

Jbern16 (approved):

PR #139, opened 2022-11-23
DA-w-Settlement/Handlers and tests

ryley-o:
## Description of the change

[IN-WORK] Add handlers for all new events associated with DA w/Refund minter, as well as tests for coverage of the new logic.

>note: ABIs are slightly updated in this PR due to a couple of added convenience functions in the refundable minter's interface to avoid subgraph bloat in future versions of refundable minters 🙏


ryley-o:
smol update to use settlement nomenclature in a56aed3

Jbern16 (approved):

PR #146, opened 2022-12-13
new minter addy's to the staging config

mchrupcala:
## Description of the change
MerkleV3 and HolderV2 contracts added for a staging (hosted) deployment.

*Note that we skipped MerkleV2 in the staging env from what I can see. Is there any reason that might cause an issue, since we do define `minterMerkleV2Contracts` in our template.yaml?

\<add description here\>

### DIFF
--- a/config/goerli-staging.json
+++ b/config/goerli-staging.json
@@ -178,11 +178,35 @@
       "startBlock": 7723588
     }
   ],
+  "minterMerkleV3Contracts": [
+    {
+      "address": "0x2FB0d2410117ac190CD6FfF017B989195A522a87",
+      "name": "Flagship Minter for V3 core",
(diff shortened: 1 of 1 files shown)


lindsgil (approved):

Jbern16:
Was there no codegen changes? I think I would expect some based on the new minters being added

mchrupcala:
> Was there no codegen changes? I think I would expect some based on the new minters being added

Hmm nope I didn't see changes after codegen was run. All my earlier changes from the dev deployment were already merged into main- do you think the type generating is already merged in from there maybe?

yoshiwarab:
Yup yup generated files were committed previously

yoshiwarab (approved):

PR #147, opened 2022-12-14
Generic Codegen Files

yoshiwarab:
## Description of the change
This PR adds a config file and scripts to package.json in order to easily create generic code-generated files (i.e. populated to folders that do include contract addresses). Previously, our flow required that we manually copy files from the generated folders into generically named folders. This PR streamlines that process. It also updates the .gitignore file so that we don't have to add every generic folder manually.

Leaving the generic files committed for now, but I think it's probably a better idea to include this generic codegen step as part of our deploy process and stop tracking any generated files.


Jbern16 (approved):

jakerockland (approved):

ryley-o:
This is awesome. Didn't realize that if address was "", this would be possible. 🙏 🙌

PR #149, opened 2022-12-20
Utilize new contract interface updates to simplify subgraph logic

yoshiwarab:
## Description of the change
Previously many of the handlers for our minter suite were for contract-specific events. This PR updates the logic to use handlers for interface events rather than contract-specific events to reduce code duplication and simplify handlers.

NOTE: I've deployed this subgraph in place of the unused Ropsten subgraph to ensure that it syncs fully without error and with the expected entities. Do not merge until it has finished syncing. https://thegraph.com/hosted-service/subgraph/artblocks/art-blocks-artist-staging?version=pending&selected=logs

### DIFF
--- a/tests/subgraph/minter-suite/minter-suite-mapping.test.ts
+++ b/tests/subgraph/minter-suite/minter-suite-mapping.test.ts
@@ -1599,7 +1599,9 @@ describe("DAExpSettlementMinters", () => {
         PROJECT_MINTER_CONFIGURATION_ENTITY_TYPE,
         getProjectMinterConfigId(minterAddress.toHexString(), project.id),
         "extraMinterDetails",
-        `{"currentSettledPrice":${actualPurchasePrice.toString()},"numSettleableInvocations":${1}}`
+        `{"currentSettledPrice":${'"' +
+          actualPurchasePrice.toString() +
+          '"'},"numSettleableInvocations":${1}}`
       );
       assert.fieldEquals(
         RECEIPT_ENTITY_TYPE,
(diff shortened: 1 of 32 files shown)


tests/subgraph/minter-suite/minter-suite-mapping.test.ts
```diff
@@ -3114,16 +3064,16 @@ describe("MinterHolder-specific tests", () => {
...
+        handleUnregisteredNFTAddress(
+          changetype<HolderUnregisteredNFTAddress>(event)
         );
       } else {
-        handleUnregisteredNFTAddressV2(
-          changetype<HolderV2UnregisteredNFTAddress>(event)
+        handleUnregisteredNFTAddress(
+          changetype<HolderUnregisteredNFTAddress>(event)
         );
       }
```
ryley-o, line 3078: Can be follow-on work, but I think we can get rid of these if statements altogether, right?

ryley-o (approved):
Thank you so much for doing this 🙏 LGTM, and a much needed improvement to our codebase!

ryley-o:
I've tagged @Jbern16 as a reviewer on this for awareness of the really great design pattern that this has unlocked 🙏

tests/subgraph/minter-suite/minter-suite-mapping.test.ts
ryley-o: resolving this because I don't want to hold up merging, and will take a look at this for follow-on activity!

yoshiwarab: Yeah we can def get rid of these, I just got lazy with the test updates

PR #150, opened 2022-12-20
Sync latest purchase price as string so we don't lose precision when we parse JSON

Jbern16:
## Description of the change

Previously, when we sync this data and parse it, we are losing precision which has ramnifications on the math used for `excess_settlement_funds`. If we sync it as string, JSON.parse() in sync should sync it to the db as as string, instead of a number

see precision loss here 
Screenshot 2022-12-20 at 10 13 51 AM

### DIFF
--- a/src/minter-suite-mapping.ts
+++ b/src/minter-suite-mapping.ts
@@ -1238,7 +1238,7 @@ function syncLatestPurchasePrice(
     projectId
   );
   // update extraMinterDetails key `currentSettledPrice` to be latestPurchasePrice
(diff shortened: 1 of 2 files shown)


yoshiwarab (commented):
🙏🙏🙏, thanks for the quick update!

yoshiwarab (approved):

lindsgil (approved):

jakerockland:
This is awesome, thank you @Jbern16 

@yoshiwarab should this be included in deploying to decentralized network?

PR #157, opened 2023-01-11
Dependency registry schema updates

yoshiwarab:
## Description of the change

\<add description here\>

>reminder: Any subgraph deployments should be documented as [Releases](https://github.com/ArtBlocks/artblocks-subgraph/releases) in this repository.

>reminder: Any changes to subgraph schema should be accompanied by corresponding changes to the Art Blocks documentation site, available at https://docs.artblocks.io/ (specifically, see the [Subgraph Entities](https://docs.artblocks.io/creator-docs/art-blocks-api/entities/) and [Subgraph Querying and API Overview](https://docs.artblocks.io/creator-docs/art-blocks-api/queries/) sections).

PR #159, opened 2023-01-11
Initial helper pass complet

yoshiwarab:
## Description of the change

\<add description here\>


src/dependency-registry.ts
```diff
@@ -0,0 +1,388 @@
...
+      return;
+    }
+    store.remove("DependencyAdditionalCDN", additionalCDN.id);
+
+    // Load each additional CDN after the removed one and decrement their index
+    for (
+      let i = event.params._additionalCDNIndex.toI32() + 1;
+      i < dependency.additionalCDNCount.toI32();
+      i++
+    ) {
```
ryley-o, line 132: Based on the smart contract's behavior, I think this has to be altered to move the dependency at the last index to `event.params._additionalCDNIndex`, instead of sliding all dependencies higher than the removed dependency down by 1.

src/dependency-registry.ts
```diff
@@ -0,0 +1,388 @@
...
+    return;
+  }
+  store.remove("DependencyAdditionalRepository", additionalRepository.id);
+
+  // Load each additional repository after the removed one and decrement their index
+  for (
+    let i = event.params._additionalRepositoryIndex.toI32() + 1;
+    i < dependency.additionalRepositoryCount.toI32();
+    i++
+  ) {
```
ryley-o, line 209: I think this has to be updated in the same way as `handleDependencyAdditionalCDNRemoved`

src/dependency-registry.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  const coreContract = IDependencyRegistryCompatibleV0.bind(
+    event.params._coreContractAddress
+  );
+
+  const scriptDetailsResult = coreContract.try_projectScriptDetails(
+    event.params._projectId
+  );
+  if (scriptDetailsResult.reverted) {
+    project.scriptTypeAndVersion = null;
```
ryley-o, line 370: just double-clicking on this - looks like we are good (on both smart contract and here) with not integrating directly with any pre-V3 core contracts, and instead requiring overrides to be set on the dependency registry 👍

ryley-o (approved):
LGTM

src/dependency-registry.ts
yoshiwarab: yupyup that's the intention

src/dependency-registry.ts
yoshiwarab: Oops wasn't thinking here! Good catch!

src/dependency-registry.ts
```diff
@@ -0,0 +1,388 @@
...
+import { DependencyRegistryV0 } from "../generated/DependencyRegistryV0/DependencyRegistryV0";
+
+export function handleDependencyAdded(event: DependencyAdded): void {
+  const dependency = new Dependency(event.params._dependencyType.toString());
+  dependency.preferredCDN = event.params._preferredCDN;
+  dependency.additionalCDNCount = BigInt.fromI32(0);
+  dependency.preferredRepository = event.params._preferredRepository;
+  dependency.additionalRepositoryCount = BigInt.fromI32(0);
+  dependency.referenceWebsite = event.params._referenceWebsite;
+  dependency.scriptCount = BigInt.fromI32(0);
```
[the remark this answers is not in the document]
yoshiwarab: Resolved in https://github.com/ArtBlocks/artblocks-subgraph/pull/167

schema.graphql
```diff
@@ -562,23 +562,35 @@ type Dependency @entity {
   additionalRepositories: [DependencyAdditionalRepository!]! @derivedFrom(field: "dependency")
   scripts: [DependencyScript!]! @derivedFrom(field: "dependency")
   scriptCount: BigInt!
+  script: String!
```
yoshiwarab, line 565: @ryley-o https://github.com/ArtBlocks/artblocks-subgraph/pull/157#discussion_r1067530159

src/dependency-registry.ts
```diff
@@ -0,0 +1,354 @@
...
+): void {
+  const coreContract = Contract.load(
+    event.params._coreContractAddress.toHexString()
+  );
+
+  if (!coreContract) {
+    return;
+  }
+
+  coreContract.dependencyRegistry = event.address;
```
yoshiwarab, line 278: @ryley-o How do you feel about this logic and the logic for the related remove function? Basically, I'm setting the dependency registry on the core contract based entirely on if its supported on the dependency registry contract. The alternative would be to create a bridge entity so that a Contract entity could be supported by multiple DependencyRegistry entities but since we're not really using that field anywhere currently I figured it wasn't worth the complexity.

ryley-o: @yoshiwarab - this feels like a fine pattern to me. It is all a little awkward because V3 core contract stores a dependency registry address and pre-V3 cores don't have a field for it, and we end up over-defined on the V3 core. We have to choose one source of truth here, and I agree that letting the dependency registry control that field in our subgraph seems most appropriate from an indexing perspective. The V3 core contract's dependency registry field is more of a pointer for someone looking at only on-chain data.

Based on the above, everything in these handlers LGTM - but you will need to go into `/src/mapping-v3-core.ts` and modify the `refreshContract` function to not assign `contractEntity.dependencyRegistry` to prevent overwriting the values assigned here!

yoshiwarab: Ah good call, thank you for the feedback!

PR #162, opened 2023-01-13
Use Interface For V3 Core Event Handlers

ryley-o:
## Description of the change

Update V3 Core to use interface events instead of contract instance events.

This is intended to be merged into main because it should have no effect on existing subgraph state.

The intent of this PR is to incorporate the interfaces changes from https://github.com/ArtBlocks/artblocks-contracts/pull/447, which will enable common event handlers for V3 flagship and V3 derivative contracts. V3 engine handlers will be added in follow-on work, that will affect existing subgraph state.


jakerockland:
@yoshiwarab gentle ping on review here as this has been sitting for a few days – could you please prioritize TAL at this this week?

yoshiwarab (approved):
Handlers for shared interfaces are so much nicer 🔥!

jakerockland (approved):
Huzzah!

PR #164, opened 2023-01-17
Remove minter types enum

### LINKED ISSUE #54: Allowlisting Minter of unknown type stops subgraph from progressing
## Current behavior (~footgun-ish):
- Subgraph stops progressing when a minter with a type not in the `enum MinterType` is allowlisted on a MinterFilter
  - somewhat easy to do if we are launching new minters and admin allowlists the new minters before a subgraph syncs

## Desired behavior:
- Subgraph logs a warning when a minter with a type not in the `enum MinterType` is allowlisted on the MinterFilter, and skips over the minter for all future operations

---
Ryley's notes on finding a solution:
- Ideally, we could alter the behavior of in `./src/helpers.ts: function loadOrCreateMinter` to validate that `minterType` is in the `enum MinterType` before attempting to save a new, invalid minter. However, I cannot find a way to get the `enum MinterType` values from `schema.graphql` in a handler function.
  - `graph codegen` command does not appear to generate anything related to the `MinterType` enum in the `./generated/` directory
- AssemblyScript does not support try/catch flows at this time: https://github.com/WebAssembly/exception-handling
- TheGraph does not support not crashing on Non-fatal errors at this time. ref: https://thegraph.com/docs/en/developer/create-subgraph-hosted/#non-fatal-errors

The only solution I can imagine would involve:
- write custom script to copy the list of enum values into our `./src/constants.ts` during `yarn prepare <...>`
  - this might be okay, but feels ...wrong
- change `Minter.type` to accept any string value
  - This would be bad for many reasons

ryley-o:
## Description of the change

This removes the minter types enum, and instead allows the subgraph to index any type of contract without immediately failing.

fixes #54 

## Benefits
- When adding new minters with new types, our subgraph won't immediately fail.

## Drawbacks
- ~a bit more likely to hit this failure mode: If we attempt to index a minter that doesn't return expected values from expected minter functions, our subgraph could fail. For example, we call `contract.minimumPriceDecayHalfLifeSeconds()` when indexing any minter who's returned type starts with `MinterDAExp`. If that function does not actually exist and the call fails, our subgraph could fail.
  - solution: we may want to harden our subgraph in follow-on PRs, for example only using try/catch patterns on all minter calls, e.g. use: `contract._minimumPriceDecayHalfLifeSeconds()`
- It is no longer obvious which minter versions are handled in our subgraph.
  - @yoshiwarab Would it be helpful to add a list of supported minters to this repo's readme? (seems like we might forget to update it?)
  
## Items not addressed in this PR:
- Ability to index new minters on-the-fly

## Design Choices
- Opted to keep minter.type as non-nullable, but it could be made nullable, e.g. in the case the call to minterType() reverts. Keeping minter.type as non-nullable keeps our schema the same and therefore does not impact downstream code, but making it nullable may be worth the effort. Cc @yoshiwarab to help weigh in on this detail.

---
  - https://app.asana.com/0/0/1203760077930845

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -321,35 +321,12 @@ type MinterFilter @entity {
   updatedAt: BigInt!
 }
 
-enum MinterType {
-  MinterSetPriceV0
-  MinterSetPriceERC20V0
-  MinterDALinV0
-  MinterDAExpV0
-  MinterSetPriceV1
-  MinterSetPriceERC20V1
-  MinterDALinV1
-  MinterDAExpV1
-  MinterHolderV0
-  MinterMerkleV0
-  MinterSetPriceV2
-  MinterSetPriceERC20V2
-  MinterDALinV2
-  MinterDAExpV2
-  MinterHolderV1
-  MinterHolderV2
-  MinterMerkleV1
-  MinterMerkleV2
-  MinterDAExpSettlementV0
-  MinterMerkleV3
-}
-
 type Minter @entity {
   "Unique identifier made up of minter contract address"
   id: ID!
 
-  "Minter type"
-  type: MinterType!
+  "Minter type - String if minter returns it's type, empty string otherwise"
+  type: String!
 
   "Associated Minter Filter"
   minterFilter: MinterFilter!
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -171,15 +171,22 @@ export function loadOrCreateMinter(
   minter.extraMinterDetails = "{}";
 
   // values assigned during contract deployments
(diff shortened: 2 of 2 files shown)


lyaunzbe (approved):

lyaunzbe:
really like the follow-on mentioned above " only using try/catch patterns on all minter calls

jakerockland (approved):
LGTM

ryley-o:
@yoshiwarab - probably worth a retroactive look by you, but merging to unblock any of the ongoing minter work by @lyaunzbe and @shantanubala (letting them avoid worrying about minter types in the enum)

PR #166, opened 2023-01-18
Add Engine Registry Handlers, Data Source Templates

ryley-o:
## Description of the change

Add Engine Registry event handlers, which dynamically add data source templates for all registered V3 core contracts not already in the subgraph store. Source template for V3 core contracts is added to our subgraph template.

ref: https://thegraph.com/docs/en/developing/creating-a-subgraph/#data-source-templates-for-dynamically-created-contracts

## Benefits
- Removes requirement to redeploy subgraph for every new engine contract.
  - Still requires adding tracked engine minter filters to the subgraph config, so for engine contracts requiring minter suite indexing, a subgraph redeployment remains necessary.

## Limitations
- Once a V3 contract is dynamically added as a data source template, it can not be removed. This means that when a contract is un-registered on the Engine Registry, the contract will remain being "watched" and update the same as previously. It will only be removed from the Engine Registry's array of registered contracts.
  - If this ends up being a bad characteristic, event handling of all Engine contracts could check the canonical Engine Registry's array of registered contracts, and only handle the event if the contract is registered (This seems overly complex to design for now, and in the event that we need to drop and Engine contract in the future, the appropriate design approach will likely become more apparent for whatever case is encountered).
- Cannot verify that contract has been successfully added as a source template in test suite
  - This capability must be verified by indexing testnet.


src/engine-registry.ts
```diff
@@ -9,11 +12,84 @@ import {
...
+  const coreAddress = event.params._contractAddress;
+  const registeredContracts = engineRegistryEntity.registeredContracts;
+  const isRegistered = indexOf(
+    registeredContracts,
+    coreAddress.toHexString()
+  ).gt(BigInt.fromI32(-1));
+  // only register the contract if it is not already registered
+  if (!isRegistered) {
+    // add the contract to the engine registry
+    registeredContracts.push(coreAddress.toHexString());
```
yoshiwarab, line 27: Since the use of array fields is generally discouraged by the Graph, we could add a registeredOn field to Contract and use that to create a relationship instead. https://thegraph.com/docs/en/developing/creating-a-subgraph/#reverse-lookups

yoshiwarab (approved):
🚀🚀🚀🚀 finally data source templates!

src/engine-registry.ts
yoshiwarab: We don't ever expect the list to grow super large so I don't feel super strongly about this though

ryley-o: @yoshiwarab - I was originally going to do it that way, but realized that a contract could potentially be registered on multiple Engine Registries. That isn't our intended use-case, but thought it might be weird if we had a new iteration of an engine registry in the future, that engine registry steals the slot on the contract, then the old engine registry doesn't list the core contract as registered in our subgraph anymore, but it is still "registered" on the blockchain contract.

Since that isn't our intended use case though, would you still prefer that we use the `registeredOn` +`@derivedFrom` pattern for this?

ryley-o: ... ah, I see there is a Map that I could use instead of an array - what do you think about that? 
https://www.assemblyscript.org/stdlib/map.html

yoshiwarab: Hmm, I'm curious if a Map could actually be serialized and stored on an entity, I would guess not. 

I think considering expected use case we could probably fudge this and just assume that it's a one to many relationship. If we don't want to do this I'm thinking that, since the length of the list of registered projects probably won't be super long at least for a while the performance hit probably won't be too drastic and the array is fine. Alternatively if we want to optimize for performance we could create a bridge entity (similar to AccountProject) to set up a many to many relationship.

I'm good with any of these honestly, so I'll leave it up to you!

ryley-o: Good call, Map wouldn't work well on the entity.

I'll go ahead and implement the `registeredOn` + `derivedFrom` pattern here since it matches our intended use cases, and remains scalable. There are also not any security implications if we forget that an engine contract is registered, since the registry is intended to be used for indexing purposes.

Thanks for all of the thoughts about this detail!

yoshiwarab: 🙏

ryley-o: @yoshiwarab - see commit b8a0f2B for updated logic and tests. I'll keep this comment thread open and let you close it it the commit looks good to you!

ryley-o:
@lyaunzbe - gentle ping on this one, it kind of wraps up the V3_Engine portion of things, and should be forward-compatible with the V3_EngineFlex 🙏

lyaunzbe (approved):
super super awesome 🔥

PR #172, opened 2023-01-24
Refactor minter handler routing to interfaces to use assumed extended interfaces

ryley-o:
## Description of the change

Refactor minter handler routing to use interface ABIs only, while assuming extended interfaces.

This removes much of the verbose and redundant handler logic having to do with adding new minter versions to the subgraph. It is accomplished by handling the most recent minter interface, and assuming that all prior versions emit events that are a subset of the most recent interface.

This was based off of some chatting with Ben Y and Ben L about how to best refactor some of our subgraph handlers to be more intentional about how to safely and more concisely add new minter versions to our subgraph 🎉


ryley-o:
THIS REMOVES 47K LINES OF CODE 🥳

yoshiwarab (approved):
Woohoo so much nicer! I think one foot gun that we might want to call out somewhere is the possibility that a future new interface may have extra functions associated with it which will be populated on the codegen'd class but not actually exist on earlier contracts so someone might accidentally assume that they can make a call that won't work.

ryley-o:
>I think one foot gun that we might want to call out somewhere is the possibility that a future new interface may have extra functions associated with it which will be populated on the codegen'd class but not actually exist on earlier contracts so someone might accidentally assume that they can make a call that won't work.

That is a great callout. I added a comment at the top of the minter-suite-mapping.ts handler file to try and help this.

PR #181, opened 2023-01-31
Bug fixes on main

ryley-o:
## Description of the change

Fix bugs that are were revealed during a subgraph sync, and clean up ABIs and generated to help prevent future bugs like these.

In an effort to reduce complexity while ensuring our `main` branch is deployable and syncable, the following updates are made:
- remove unused ABIs
- archive all secondary sales subgraph handlers, tests, and templates
- add necessary ABIs to the subgraph manifest
  - Note: missing ABIs are currently only caught during a subgraph sync, so an issue has been raised to The Graph team about this quirk
- discontinue tracking `/genrated/` directory in github
  - this directory is not needed for CI or github, and can lead to a stale state that may hide issues
  - this directory often clutters up our PRs
- bug fix to an EngineRegistry handler

Note that I've done my best to determine the minimum set of ABIs required in the subgraph template yaml file. This is not brute-force include-everything, but it still might be a slight over-estimate.


config/goerli-staging.json
```diff
@@ -14,13 +14,6 @@
       "startBlock": 7723578
     }
   ],
-  "genArt721CoreContracts": [
-    {
-      "address": "0xDa62f67BE7194775A75BE91CBF9FEeDcC5776D4b",
-      "name": "Artist Staging Goerli Flagship Core V1 Contract",
-      "startBlock": 7011002
-    }
-  ],
```
ryley-o, line 23: this was a duplicate key and entry

yoshiwarab (approved):
Thank you! I think it would make sense to update the codegen command in package.json to first remove the generated directory before regenerating it. Also, since these are no longer tracked we should add an extra step to all the deploy commands that does codegen first.

jakerockland (approved):

PR #183, opened 2023-01-31
Add Admin ACL data source templates

ryley-o:
## Description of the change

This adds Admin ACL Data source templates so we don't require a subgraph redeployment for every new Engine contract being indexed.

Admin ACL contracts are required to be indexed to handle the case where the superAdmin address is changed.

This update also updates the dev config to remove the adminACL contracts currently deployed to dev (since they will now be automatically added via data source templates in the Engine Registry handler).


jakerockland (approved):
🙏

yoshiwarab (approved):

PR #184, opened 2023-01-31
Update dev and staging configs

ryley-o:
## Description of the change

Shuffle which contracts are being indexed on dev and staging, and add any new dev or staging minters we want to index for the new V3 Engine contracts.

Also updates mainnet config to include ABxPace, ABxBMG, and the AB managed Engine contract. Minters are included for ABxPace and ABxBMG.

This PR is intended to wrap up our recent set of subgraph changes and deployments, and should be used for releases on all testnet and mainnet subgraphs. 🥳

### DIFF
--- a/config/goerli-dev.json
+++ b/config/goerli-dev.json
@@ -81,16 +81,6 @@
       "startBlock": 7799145
     },
     {
-      "address": "0x5E22FBBB6bb2C45aE060b4C96f5039efdF7Fb328",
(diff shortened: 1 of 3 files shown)


jakerockland (approved):
This LGTM ser.

jakerockland:
This is purely a config change so I think OK to merge with single review given that we're going first to goerli before doing any mainnet 🚢

Issue #108, opened 2022-09-23
Improve subgraph local deployment testing


ryley-o:
This repository's testing is great, but additional testing associated with running a subgraph locally can catch many other important bugs. For example, the contents of `subgraph.yaml` can only be tested in this manner, since testing in matchstick requires us to route events to their appropriate handlers directly.

We already have scripts to test locally (e.g. `yarn deploy:local` can be used to deploy to local graph node), but defining and documenting some general scripted commands on a contract suite would make sense. See the discussion in this PR in our smart-contracts repo: https://github.com/ArtBlocks/artblocks-contracts/pull/302#issuecomment-1256463269

Ideally, we could incorporate something like what is described in the link above to streamline and document local deployments of subgraphs.

PR #193, opened 2023-03-01
E2E test setup

yoshiwarab:
## Description of the change

Initial set up of an e2e test framework using docker compose.


tests/e2e/hardhat/hardhat.config.js
```diff
@@ -0,0 +1,20 @@
+require("@nomiclabs/hardhat-ethers");
+const fs = require("fs");
+
+task('node', 'Starts a Hardhat node', async (args, hre, runSuper) => {
+  const accounts = config.networks.hardhat.accounts;
```
yoshiwarab, line 5: This is to save the default hardhat network accounts mnemonic to a shared volume where it can be read to create a wallet for contract interactions. 

Alternatively we could have an env file that specifies the mnemonic that gets read anywhere we need accounts to use.

package.json
```diff
@@ -4,6 +4,7 @@
   "scripts": {
     "build": "yarn codegen:generic && graph build",
     "test": "yarn codegen:generic && graph test -r",
+    "test:e2e": "cd tests/e2e && docker compose up --exit-code-from runner",
```
yoshiwarab, line 7: This will cause the docker compose up command to exit with whatever code the runner container exits with. We can take advantage of this to determine if our tests pass in CI.

Additionally, we could expose relevant ports in the docker compose and run docker compose up without this flag so that we can interact with this setup locally. Definitely a fast follow on this work.

ryley-o:
@yoshiwarab - I'm having an issue getting this to run - let's troubleshoot tomorrow!

ryley-o (approved):
This is brilliant 👑 thank you, and this will add so much robustness to our subgraph deployments 🙏

tests/e2e/hardhat/hardhat.config.js
jakerockland: Awesome–could we call this first part out in an inline comment for posterity?

package.json
jakerockland: 🔥 🤌

jakerockland (approved):

ryley-o:
@yoshiwarab - I think this closes #108, yeah?

PR #194, opened 2023-03-02
Add serial English auction minter

ryley-o:
## Description of the change

Add support for indexing a new serial English Auction (SEA) minter.

This PR includes all updates and tests associated with indexing the new minter added in: https://github.com/ArtBlocks/artblocks-contracts/pull/504

No schema changes are required. This is due to using existing fields and generic json fields to store all new minter parameters.

---
  - https://app.asana.com/0/0/1203665500354291

### DIFF
--- a/abis/IFilteredMinterSEAV0.json
+++ b/abis/IFilteredMinterSEAV0.json
@@ -1,0 +1,953 @@
+{
+  "_format": "hh-sol-artifact-1",
+  "contractName": "IFilteredMinterSEAV0",
+  "sourceName": "contracts/interfaces/0.8.x/IFilteredMinterSEAV0.sol",
+  "abi": [
+    {
+      "anonymous": false,
+      "inputs": [
+        {
+          "indexed": true,
+          "internalType": "uint256",
+          "name": "tokenId",
+          "type": "uint256"
+        },
+        {
+          "indexed": true,
+          "internalType": "address",
+          "name": "bidder",
+          "type": "address"
+        },
+        {
+          "indexed": false,
+          "internalType": "uint256",
+          "name": "bidAmount",
+          "type": "uint256"
+        }
+      ],
+      "name": "AuctionBid",
+      "type": "event"
+    },
+    {
+      "anonymous": false,
+      "inputs": [
+        {
+          "indexed": false,
+          "internalType": "uint32",
+          "name": "minAuctionDurationSeconds",
+          "type": "uint32"
+        },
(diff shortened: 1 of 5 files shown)


src/minter-suite-mapping.ts
```diff
@@ -492,6 +505,255 @@ export function handleDelegationRegistryUpdatedGeneric<T>(event: T): void {
   minter.save();
 }
 
+// MinterSEA specific handlers
+export function handleAuctionDurationSecondsRangeUpdated(
+  event: AuctionDurationSecondsRangeUpdated
+): void {
+  let minter = loadOrCreateMinter(event.address, event.block.timestamp);
+
+  if (!minter) {
```
jakerockland, line 514: In this case here, and below, should we log a warning within these guards?

If we do that, should we add a utility function?

abis/IFilteredMinterSEAV0.json
```diff
@@ -0,0 +1,972 @@
+{
```
jakerockland, line 1: Should we update this repo to import the ABI npm package so that we don't need to manually import like this?

jakerockland (approved):

abis/IFilteredMinterSEAV0.json
ryley-o: This is a good callout. I would like to do this in a follow-on PR to isolate the change, and have added an issue for this here: https://github.com/ArtBlocks/artblocks-subgraph/issues/197

jakerockland: SGTM!

jakerockland: Thank you!! :)

yoshiwarab (approved):

src/minter-suite-mapping.ts
ryley-o: good point - realized we don't even need these null minter guards because `loadOrCreateMinter` never returns null. This was some old habits from when we didn't have the helper function. Updated in 061dd92 🙏 

Note that if we do have guards like these, in general, I would prefer to log a warning to help keep us aware of any ongoing issues in deployed subgraphs!

ryley-o:
Tagging this with `do not merge` until smart contract repo's PR is merged 🙏

ryley-o:
Oops I guess the do not merge tag still merges 😓 this is fine, and any updates from any smart contract iteration will be made in subsequent PRs.

PR #195, opened 2023-03-06
V3 Engine Flex Updates

lyaunzbe:
## Description of the change

Integrating support & indexing for newly added V3 Engine Flex contracts:
- Adds new dependency type to flex dependecy type enums ("ONCHAIN")
- ProjectExternalAssetDependency entity adds data and bytecode nullable fields (popualted when dependency type is "ONCHAIN")
- V3 Engine Flex tests

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -580,7 +580,8 @@ type Dependency @entity {
   "Additional repositories for this dependency"
   additionalRepositoryCount: BigInt!
   "Number of additional repositories for this dependency"
-  additionalRepositories: [DependencyAdditionalRepository!]! @derivedFrom(field: "dependency")
+  additionalRepositories: [DependencyAdditionalRepository!]!
+    @derivedFrom(field: "dependency")
   "List of on-chain scripts that for this dependency"
   scripts: [DependencyScript!]! @derivedFrom(field: "dependency")
   "Number of on-chain scripts for this dependency"
@@ -648,6 +649,8 @@ enum ProjectExternalAssetDependencyType {
   IPFS
   "Asset hosted on Arweave"
   ARWEAVE
+  "Asset hosted on chain"
+  ONCHAIN
 }
 
 type ProjectExternalAssetDependency @entity {
@@ -665,4 +668,10 @@ type ProjectExternalAssetDependency @entity {
 
   "The dependency index"
   index: BigInt!
+
+  "Address of the bytecode storage contract for this asset if it is onchain"
+  bytecodeAddress: Bytes
+
+  "The asset data if it is onchain"
+  data: String
 }
--- a/schema.graphql
+++ b/schema.graphql
@@ -663,13 +663,13 @@ type ProjectExternalAssetDependency @entity {
   "The dependency type"
   dependencyType: ProjectExternalAssetDependencyType!
 
-  "The dependency cid"
+  "The dependency cid. This will be an empty string assets of type ONCHAIN."
   cid: String!
 
   "The dependency index"
   index: BigInt!
 
-  "Address of the bytecode storage contract for this asset if it is onchain"
+  "Address of the bytecode storage contract for this asset if it is of type ONCHAIN"
   bytecodeAddress: Bytes
 
   "The asset data if it is onchain"
--- a/src/engine-registry.ts
+++ b/src/engine-registry.ts
@@ -34,7 +34,7 @@ export function handleContractRegistered(event: ContractRegistered): void {
     IGenArt721CoreV3_Base_Template.create(coreAddress);
     OwnableGenArt721CoreV3Contract_Template.create(coreAddress);
     IERC721GenArt721CoreV3Contract_Template.create(coreAddress);
-    // okay to create this template even if the contract is not engine flex as event handlers do not overlap
+    // @dev okay to create this template even if the contract is not engine flex as event handlers do not overlap
     IGenArt721CoreContractV3_Engine_Flex_Template.create(coreAddress);
     // also track the new contract's Admin ACL contract to enable indexing if admin changes
     // @dev for V3 core contracts, the admin acl contract is the core contract's owner
--- a/src/engine-registry.ts
+++ b/src/engine-registry.ts
@@ -7,7 +7,8 @@ import {
   IGenArt721CoreV3_Base_Template,
   OwnableGenArt721CoreV3Contract_Template,
   IERC721GenArt721CoreV3Contract_Template,
-  AdminACLV0_Template
+  AdminACLV0_Template,
+  IGenArt721CoreContractV3_Engine_Flex_Template
 } from "../generated/templates";
 
 import {
@@ -33,6 +34,7 @@ export function handleContractRegistered(event: ContractRegistered): void {
     IGenArt721CoreV3_Base_Template.create(coreAddress);
     OwnableGenArt721CoreV3Contract_Template.create(coreAddress);
     IERC721GenArt721CoreV3Contract_Template.create(coreAddress);
+    IGenArt721CoreContractV3_Engine_Flex_Template.create(coreAddress);
     // also track the new contract's Admin ACL contract to enable indexing if admin changes
     // @dev for V3 core contracts, the admin acl contract is the core contract's owner
     const ownableV3Core = Ownable.bind(coreAddress);
--- a/src/engine-registry.ts
+++ b/src/engine-registry.ts
@@ -34,6 +34,7 @@ export function handleContractRegistered(event: ContractRegistered): void {
     IGenArt721CoreV3_Base_Template.create(coreAddress);
     OwnableGenArt721CoreV3Contract_Template.create(coreAddress);
     IERC721GenArt721CoreV3Contract_Template.create(coreAddress);
+    // okay to create this template even if the contract is not engine flex as event handlers do not overlap
     IGenArt721CoreContractV3_Engine_Flex_Template.create(coreAddress);
     // also track the new contract's Admin ACL contract to enable indexing if admin changes
     // @dev for V3 core contracts, the admin acl contract is the core contract's owner
--- a/src/mapping-v3-core.ts
+++ b/src/mapping-v3-core.ts
@@ -983,7 +983,7 @@ function _handleExternalAssetDependencyUpdated<T>(
 }
 
 export function handleExternalAssetDependencyRemoved(
-  event: ExternalAssetDependencyUpdated
+  event: ExternalAssetDependencyRemoved
 ): void {
   const engineFlexContract = getV3EngineFlexContract(event.address);
   if (engineFlexContract) {
--- a/src/mapping-v3-core.ts
+++ b/src/mapping-v3-core.ts
@@ -964,13 +964,13 @@ function _handleExternalAssetDependencyUpdated<T>(
     FLEX_CONTRACT_EXTERNAL_ASSET_DEP_TYPES[event.params._dependencyType];
 
   if (assetEntity.dependencyType === "ONCHAIN") {
-    const projextExternalAssetDependency = contract.try_projectExternalAssetDependencyByIndex(
+    const projextExternalAssetDependency = contract.projectExternalAssetDependencyByIndex(
       event.params._projectId,
       event.params._index
     );
     assetEntity.bytecodeAddress =
-      projextExternalAssetDependency.value.bytecodeAddress;
-    assetEntity.data = projextExternalAssetDependency.value.data;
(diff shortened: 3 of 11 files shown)


jakerockland (approved):

src/mapping-v3-core.ts
```diff
@@ -892,6 +919,225 @@ function _handleOwnershipTransferred<T>(
...
+    assetEntity.index = index;
+    assetEntity.dependencyType =
+      FLEX_CONTRACT_EXTERNAL_ASSET_DEP_TYPES[
+        contractExternalAsset.value.dependencyType
+      ];
+    if (assetEntity.dependencyType === "ONCHAIN") {
+      assetEntity.bytecodeAddress = contractExternalAsset.value.bytecodeAddress;
+      assetEntity.data = contractExternalAsset.value.data;
+    }
+    assetEntity.save();
```
ryley-o, line 1063: I think we need a `if (!projextExternalAssetDependency.reverted) {` enclosing these (due to the use of `try_`).

src/mapping-v3-core.ts
```diff
@@ -892,6 +919,225 @@ function _handleOwnershipTransferred<T>(
...
+  }
+
+  project.externalAssetDependencyCount = lastEntityIndex;
```
ryley-o, line 1066: nit - maybe a dev comment here explaining the last index is the previous count minus 1?

src/mapping-v3-core.ts
```diff
@@ -1047,7 +1309,7 @@ function refreshContract<T>(contract: T, timestamp: BigInt): Contract | null {
     contractEntity.curationRegistry = contract.artblocksCurationRegistryAddress();
     // flagship never auto approves artist split proposals for all changes
     contractEntity.autoApproveArtistSplitProposals = false;
-  } else if (contract instanceof GenArt721CoreV3_Engine) {
+  } else if (isEngineOrEngineFlex) {
```
ryley-o, line 1312: nit - WDYT about a comment somewhere in this code block stating that all flex-specific functionality is not refreshed, due to the flex-specific handlers updating relevant entities on their own?

subgraph.template.yaml
```diff
@@ -2,6 +2,74 @@ specVersion: 0.0.2
...
+        - name: IGenArt721CoreContractV3_Base
+          file: ./abis/IGenArt721CoreContractV3_Base.json
+        - name: IGenArt721CoreContractV3
+          file: ./abis/IGenArt721CoreContractV3.json
+        - name: IGenArt721CoreContractV3_Engine
+          file: ./abis/IGenArt721CoreContractV3_Engine.json
+        - name: IGenArt721CoreContractV3_Engine_Flex
+          file: ./abis/IGenArt721CoreContractV3_Engine_Flex.json
+        - name: GenArt721CoreV3_Engine_Flex
+          file: ./abis/GenArt721CoreV3_Engine_Flex.json
```
ryley-o, line 39: you _might_ need to add these ABIs to the other V3 core dataSources. I'm not 100% sure, but I do know missing ABIs is not caught until runtime.

subgraph.template.yaml
```diff
@@ -1636,3 +1704,68 @@ templates:
...
```
ryley-o, line 1739: (same comment about missing ABIs in the other data source templates)

src/mapping-v3-core.ts
```diff
@@ -93,7 +106,8 @@ function _handleMint<T>(contract: T, event: Mint): void {
   if (
```
ryley-o, line 106: I think every section before these type of checks needs an additional check:
```
const engineFlexContract = getV3EngineFlexContract(event.address);
  if (engineFlexContract) {
    _handleMint(engineContract, event);
    return;
  }
```

ryley-o (commented):
Took an initial look and have some feedback, couple minor nits. In addition to the comments below:

- in `/src/engine-registry.ts`, ~line 36 - if new contract entity is an engine flex, need to create a new data source template
- for additional test coverage, I _think_ you could copy-paste `mapping-v3-engine-core.test.ts`, and change line `92` to be `const coreType = "GenArt721CoreV3_Engine_Flex";`. If I'm remembering how those tests are set up, that would allow the functionality that is common between Engine and Engine Flex to be tested/confirmed

src/mapping-v3-core.ts
lyaunzbe: ah roger, but I think this can just be added once to the upper level `handleMint` function if im understanding correctly

src/mapping-v3-core.ts
```diff
@@ -892,6 +919,225 @@ function _handleOwnershipTransferred<T>(
...
+    FLEX_CONTRACT_EXTERNAL_ASSET_DEP_TYPES[event.params._dependencyType];
+
+  if (assetEntity.dependencyType === "ONCHAIN") {
+    const projextExternalAssetDependency = contract.try_projectExternalAssetDependencyByIndex(
+      event.params._projectId,
+      event.params._index
+    );
+    assetEntity.bytecodeAddress =
+      projextExternalAssetDependency.value.bytecodeAddress;
+    assetEntity.data = projextExternalAssetDependency.value.data;
```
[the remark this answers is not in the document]
lyaunzbe: I'm fine with adding the `.reverted` check personally (and think it will be easier at this point). but also, thinking more on it we may just be able to use a regular contract call vs 'try_'. The only way this contract call reverts is if the external asset dependency, which we just received an event for in this handler, wouldn't exist (and this shouldn't be possible).

schema.graphql
```diff
@@ -648,6 +649,8 @@ enum ProjectExternalAssetDependencyType {
   IPFS
   "Asset hosted on Arweave"
   ARWEAVE
+  "Asset hosted on chain"
+  ONCHAIN
 }
 
 type ProjectExternalAssetDependency @entity {
```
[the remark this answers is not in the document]
lyaunzbe: good call! will add that comment

src/mapping-v3-core.ts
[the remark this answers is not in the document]
ryley-o: Agree, and we certainly do not rely on checking for reverted in other handlers. At some point we should codify which is preferred in this repo, but for now seems like we either remove the `try_`, or add the `!reverted` check?

src/mapping-v3-core.ts
ryley-o: yes ser! sorry, Github being a bit weird about where I could place this comment 😓

yoshiwarab: I think the `tokenIdToHash` call is the only thing that's preventing us from having a single handler here that uses `IGenArt721CoreContractV3_Base.bind(contractAddress)` for everything so that we don't need to do the whole contract type check pattern. I think that's common across all of the V3 contracts so it would make sense to me to add to the interface.

I'd imagine a lot of the other handlers for events exported from `../generated/IGenArt721CoreV3_Base/IGenArt721CoreContractV3_Base` could be updated similarly?

tests/subgraph/mapping-v3-core/mapping-v3-engine-flex-core.test.ts
```diff
@@ -0,0 +1,975 @@
...
+  addNewProjectToStore(
+    TEST_CONTRACT_ADDRESS,
+    projectId,
+    "Test Project",
+    randomAddressGenerator.generateRandomAddress(),
+    BigInt.fromI64(i64(1e18)),
+    CURRENT_BLOCK_TIMESTAMP.minus(BigInt.fromI32(100))
+  );
+
+  const event: ExternalAssetDependencyUpdated = changetype<
```
yoshiwarab, line 132: Super minor nit but would be helpful when reading this test if this event was called updateEvent1 or creationEvent given the context of the other 2 events

tests/subgraph/mapping-v3-core/mapping-v3-engine-flex-core.test.ts
```diff
@@ -0,0 +1,975 @@
...
+    fullProjectId + "-" + _index0.toString(),
+    "project",
+    fullProjectId
+  );
+  
+  const updateEvent: ExternalAssetDependencyUpdated = changetype<
+    ExternalAssetDependencyUpdated
+  >(newMockEvent());
+  updateEvent.address = TEST_CONTRACT_ADDRESS;
+  updateEvent.block.timestamp = CURRENT_BLOCK_TIMESTAMP;
```
yoshiwarab, line 211: It would be helpful to choose another later timestamp so that we can assert that it has changed.

tests/subgraph/mapping-v3-core/mapping-v3-engine-flex-core.test.ts
```diff
@@ -0,0 +1,975 @@
...
+    fullProjectId,
+    "externalAssetDependencyCount",
+    _externalAssetDependencyCount1.toString()
+  );
+  
+  const updateEvent2: ExternalAssetDependencyUpdated = changetype<
+    ExternalAssetDependencyUpdated
+  >(newMockEvent());
+  updateEvent2.address = TEST_CONTRACT_ADDRESS;
+  updateEvent2.block.timestamp = CURRENT_BLOCK_TIMESTAMP;
```
yoshiwarab, line 263: Same timestamp comment as above

subgraph.template.yaml
yoshiwarab: yup, discussed with Ben offline but wanted to comment here for posterity. Everywhere we're doing the 
```
const flagshipContract = getV3FlagshipContract(event.address);
  if (flagshipContract) {
    ...
  }
  const engineContract = getV3EngineContract(event.address);
  if (engineContract) {
    ...
  }
```
we're potentially calling the GenArt721CoreV3_Engine_Flex contract so we need the abi. I think in practice we'd early return everywhere before hitting this unless it was an unknown contract but i think since it's potentially called in the same handler it makes sense to add the abi. Good catch!

src/mapping-v3-core.ts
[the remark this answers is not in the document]
lyaunzbe: went with removing `try_` 👍

src/mapping-v3-core.ts
ryley-o: @yoshiwarab ah, yes, good call, agree for the case of this `handleMint` handler! This might also be true for a few of the other handlers, and seems best to rely on `IGenArt721CoreContractV3_Base.bind` where possible and straightforward. For things like `handlePlatformUpdated` I _think_ the current pattern of binding to the contract type is just clean because it enables statements like `if (contract instanceof GenArt721CoreV3_Engine)` in areas where the handling behaviors diverge between the engine and flagship (e.g. when dealing with payment addresses or indexing of minter suite).

yoshiwarab: Cool, yeah makes sense! Also, I think this more general refactor should be a follow on rather than part of this PR

src/engine-registry.ts
```diff
@@ -33,6 +34,7 @@ export function handleContractRegistered(event: ContractRegistered): void {
     IGenArt721CoreV3_Base_Template.create(coreAddress);
     OwnableGenArt721CoreV3Contract_Template.create(coreAddress);
     IERC721GenArt721CoreV3Contract_Template.create(coreAddress);
+    IGenArt721CoreContractV3_Engine_Flex_Template.create(coreAddress);
```
ryley-o, line 37: We could conditionally create this flex data source template only in the case where the contract is of type flex. I don't have any super strong opinions here other than if we keep it as-is, probably worth a dev comment indicating why it is okay to create a flex template for non-flex contracts (i.e. the event handlers don't overlap)

subgraph.template.yaml
```diff
@@ -1636,3 +1732,68 @@ templates:
         - event: SuperAdminTransferred(indexed address,indexed address,address[])
           handler: handleIAdminACLV0SuperAdminTransferred
       file: ./src/mapping-v3-core.ts
+
+  - name: IGenArt721CoreContractV3_Engine_Flex_Template
```
ryley-o, line 1736: optional minor nit: can we move this above the AdminACLV0_Template, just to keep the core contract templates all together?

ryley-o (approved):
LGTM! 🎉

src/mapping-v3-core.ts
ryley-o: @lyaunzbe - I'm happy to pick up this refactoring work in a fast-follow PR if that sounds good to you? 👍

lyaunzbe: @ryley-o yes, that would be much appreciated! thank you ser 🙏 🙏

PR #212, opened 2023-03-30
Move minter specific fields to extra

ryley-o:
## Description of the change

Move fields specific to certain minters into `extraMinterDetails` on `Minter` and `ProjectMinterConfiguration` entities.

Internal mini-design doc for this: https://docs.google.com/document/d/1zotH39ijFw06vw4nXkyOVNLBwLIcbn582R77_QIN6h8/edit?usp=sharing

This is backwards-compatible to any consumers of this information, and labels minter-specific fields as deprecated.

## Design Chocies
- Given that it is not a backwards compatible schema, and we would like to support a smooth migration on Flagship as well as for Engine partners, this change labels minter-specific fields as deprecated, but keeps them populated while this migration is impemented.

## Follow-on

After new subgraphs are deployed with this change, a full project sync will be required. Also note that grafted subgraph deployments will not properly capture this change, since it affects historical indexing behaviors.

---
  - https://app.asana.com/0/0/1204293293986911


src/helpers.ts
```diff
@@ -223,7 +239,9 @@ export function getMinterDetails<T>(config: T): TypedMap<string, JSONValue> {
   if (
     !(config instanceof ProjectMinterConfiguration || config instanceof Minter)
   ) {
-    throw new Error("cannot find extra minter details on " + config);
```
ryley-o, line 226: this is weird, but for some reason the changes in this PR no longer allow the previous code. Kind of a code smell, but I can't figure out why this needed to be changed...

jakerockland: Does this work? 

```
`cannot find extra minter details on {config}`
```

ryley-o: assuming you meant with the dollar sign, it doesn't work:
```
`cannot find extra minter details on ${config}`
```

I think this is actually because of some change in ordering of `handleSetMinterDetailsGeneric`, which improves our robustness by not calling this function with a generic type. This probably would have crashed the subgraph before if we had sent over something that was not a `ProjectMinterConfiguration` or `Minter`, and that didn't have a `toString()` method 💡

ryley-o:
@jakerockland @yoshiwarab Gentle ping on this one, as we will eventually want this folded in before switching to the new minter suite architecture. Cc @lyaunzbe

jakerockland:
> @jakerockland @yoshiwarab Gentle ping on this one, as we will eventually want this folded in before switching to the new minter suite architecture. Cc @lyaunzbe

Gah! Apologies for the delay and TYVM for the extra ping here. Will review today! 🙏

jakerockland (approved):

yoshiwarab (approved):

PR #214, opened 2023-04-14
Phase I.A. Update schema for new minter suite architecture

ryley-o:
## Description of the change

Update subgraph schema for new minter suite architecture.

These changes reflect an updated architecture where a single minter filter supports many core contracts, instead of a single core contract.

## Next Steps

Because these are breaking changes, these changes are proposed to be merged into a feature branch, `minter-suite-v2`.

Follow-up steps are:
- update existing handlers to conform to this new schema
- add new handlers for new shared minter filter and engine registry
- add new handlers for new shared minters

Associated design doc: https://docs.google.com/document/d/1Ix0LhPHqeDgCouKPSYZufQY9zQ5cNqf8yXHbus8y3w8/edit?usp=sharing

---
associated asana task: https://app.asana.com/0/0/1204519240119681/f

>note: this is built off of the refactor-json branch because that strategy will be merged into main before this feature is completed.


ryley-o:
removed two legacy-only fields, based on conversation w/ @yoshiwarab 🙏 [cbd4714](https://github.com/ArtBlocks/artblocks-subgraph/pull/214/commits/cbd47147d2cf073102a001f4aba6cf8013c15f04)

schema.graphql
```diff
@@ -171,11 +171,11 @@ type ProposedArtistAddressesAndSplit @entity {
   createdAt: BigInt!
 }
 
-type EngineRegistry @entity {
-  "Unique identifier made up of the Engine Registry's contract address"
+type CoreRegistry @entity {
+  "Unique identifier made up of the Core Registry's contract address (legacy name was Engine Registry). Note: For legacy MinterFilters, this is a dummy ID, equal to the address of the single core contract associated with the minter filter."
```
shantanubala, line 175: Tiny nit: this would be more readable as a multi-line comment delimited by `"""`

schema.graphql
```diff
@@ -343,14 +343,34 @@ type MinterFilter @entity {
...
   id: ID!
 
-  "Associated core contract"
-  coreContract: Contract!
+  "Minters allowlisted globally on this MinterFilter"
+  minterGlobalAllowlist: [Minter!]!
 
-  "Minters allowlisted on MinterFilter"
-  minterAllowlist: [Minter!]!
+  "Minters allowlisted at a contract-level on this MinterFilter, only used for MinterFilterV2+"
```
shantanubala, line 349: Also tiny nit: I'd like this comment to describe more specifically that this allowlist extends the global allowlist (minter must exist on global OR contract allowlist, and not expected to exist in both, right?).

schema.graphql
```diff
@@ -343,14 +343,34 @@ type MinterFilter @entity {
...
+  id: ID!
+
+  "MinterFilter contract"
+  minterFilter: MinterFilter!
+
+  "Core contract"
+  contract: Contract!
+
+  "Minter contract addresses allowed at the contract level (extending global MinterFilter allowlist)"
+  minterContractAllowlist: [Bytes!]!
```
shantanubala, line 373: Is there a reason this is a `Bytes` array while the global allowlist is a `Minter` array?

Is this because we expect engine partners to use minters that we haven't seen before / aren't indexing ourselves, but should still be allowed to mint?

schema.graphql
```diff
@@ -343,14 +343,34 @@ type MinterFilter @entity {
...
+  "Minters allowlisted at a contract-level on this MinterFilter, only used for MinterFilterV2+"
+  minterFilterContractAllowlists: [MinterFilterContractAllowlist!]!
+    @derivedFrom(field: "minterFilter")
 
-  "Minters associated with MinterFilter"
-  associatedMinters: [Minter!]! @derivedFrom(field: "minterFilter")
+  "Known minters that are tied to this MinterFilter, but are not necessarily approved on this MinterFilter"
+  knownMinters: [Minter!]! @derivedFrom(field: "minterFilter")
+
+  "Core contract registry used by this MinterFilter. Note that for MinterFilterV0 & MinterFilterV1, a dummy CoreRegistry is created in the subgraph that contains a single allowlisted core contract."
```
shantanubala, line 356: Tiny nit: Also a good candidate for  `"""` multi-line comments

shantanubala (approved):
LGTM with some really minor nits where it took me a little longer than I'd have liked to understand what was going on.

schema.graphql
ryley-o: ah, thank you, good call!

schema.graphql
ryley-o: I've updated this to indicate that this enables a core contract to extend the global minter allowlist to allow minters that are not globally allowlisted.
(in general, not expected that a minter would be on this list and the global list, but nothing explicitly prevents that)

schema.graphql
ryley-o: good catch - I think this actually should be an array of `[Minter!]!`, so I've update.
We do expect engine partners to use minters that we haven't seen before/aren't indexing, but I think we would still generically be fine creating Minter entities in the store from them, and therefore should have an array of `Minter`s here.

Do want to note that an array here is certainly a design choice, and we expect this specific array to remain relatively short. We could make this a "derived from" field if we added yet another entity `MinterFilterContractAllowlisting`, that defined each individual minter-core-minterfilter allowlisting...

ryley-o:
gentle ping here @yoshiwarab @lyaunzbe 🙏

schema.graphql
```diff
@@ -171,11 +171,15 @@ type ProposedArtistAddressesAndSplit @entity {
...
+type CoreRegistry @entity {
+  """
+  Unique identifier made up of the Core Registry's contract address (legacy name was Engine Registry).
+  Note: For legacy MinterFilters, this is a dummy ID, equal to the address of the single core contract
+  associated with the minter filter.
+  """
   id: ID!
 
-  "Core contracts that are registered on this Engine Registry, when this is most recent Engine Registry to add the contract"
+  "All core contracts that are registered on this CoreRegistry, when this is most recent Core Registry to add the contract"
```
yoshiwarab, line 182: Just curious, do we ever expect a contract would be registered on more than one registry?

yoshiwarab (approved):

schema.graphql
ryley-o: I don't think so, at least on in a meaningful way where we would want to know anything but the most recent core registry.
I'm going to resolve this, but @jakerockland feel free to chime in if you have any additional thoughts about this.

schema.graphql
ryley-o: resolving this since it has been updated, and the design choice noted in the internal design doc.

PR #223, opened 2023-05-24
Phase II.A: New minter filter handlers

ryley-o:
## Description of the change

Add handlers for new shared minter filter.

Robustness of new and legacy minter filter handler events are also generally improved by adding `try_` to all calls to the minter filter contract (and some other contracts).

## Testing
All new tests for the new shared minter filter are written as e2e tests. A few comments on e2e tests:
- Seems like a much better overall testing strategy than matchstick due to reduced mocks, manually adding things to the store, ability to now catch missing ABIs in subgraph config, etc.
- Some edge cases are not easily testable (or even not testable at all with our current set of contracts) due to on-chain constraints
  - e.g. cannot test redundant check in our subgraph for if a minter being approved twice on minter filter, because smart contract already disallows that.
  - This generally seems fine, because these cases should not be able to be encountered in real life, and our subgraph guards are typically very straightforward.
- No coverage reports are generated for e2e testing (coverage reports were rarely used and low quality in Matchstick)
- e2e tests do take longer to run, but test code is much more straightforward and easy to read/write (both conceptually, and because we are now avoiding AssemblyScript)

⚠️ We do NOT currently have CI set up to verify that e2e tests are passing prior to the PR merging ⚠️ 

TODO

- [x] Add initial cut of template + handler logic
- [x] Update minterFilterUpdated logic in V3 core handers
- [x] Add e2e tests for coverage
- [x] Update to new contracts npm release that includes `IMinterFilterV1.sol`

---
  - https://app.asana.com/0/0/1204661047384538


src/helpers.ts
```diff
@@ -184,9 +252,23 @@ export function loadOrCreateMinter(
   let filteredMinterContract = IFilteredMinterV2.bind(minterAddress);
 
   // values assigned in contract constructors
-  minter.minterFilter = filteredMinterContract
-    .minterFilterAddress()
-    .toHexString();
+  // @dev safely retrieve value, gracefully handle if it reverts
```
ryley-o, line 255: note: important to add safety here, because Engine contracts can allowlist any minter, which may or may not conform to the expected interfaces

tests/subgraph/mapping-v3-core/helpers.ts
```diff
@@ -272,6 +275,16 @@ export function mockMinterUpdatedCallsNoPreconfiguredProjects(
...
+
+  // should handle unsupported minterFilterType function, since some legacy
+  // minter filters (including flagship MinterFilterV1) don't implement the
+  // function
+  // @dev shared minter types not needed, since e2e testing is used for those minters
+  createMockedFunction(
+    TEST_MINTER_FILTER_ADDRESS,
+    "minterFilterType",
+    "minterFilterType():(string)"
+  ).reverts();
```
ryley-o, line 287: Although the current MinterFilterV1 in the contracts repo does implement a minterFilterType function, the originally deployed MinterFilterV1 used by our flagship V3 core did not implement the function. Therefore, we should handle the revert case gracefully.

src/shared-minter-filter-mapping.ts
```diff
@@ -0,0 +1,335 @@
...
+    log.warning(
+      "[WARN] Project minter at {} does not match minter filter that emitted the ProjectMinterRegistered event at {}",
+      [minter.minterFilter, minterFilter.id]
+    );
+  }
+
+  // if the project's minter filter does not match the minter filter that
+  // emitted the ProjectMinterRemoved event, then this is a pre-configuring
+  // event, and we should return early and not update the project's minter
+  const coreContract = Contract.load(event.params.coreContract.toHexString());
```
yoshiwarab, line 216: Nit: Comments definitely make this clear, but might be slightly easier to read at first glance if we assign coreContract.minterFilter to a descriptive variable like currentProjectMinterFilter or something

tests/e2e/runner/.dockerignore
```diff
@@ -1,2 +1,3 @@
 node_modules
-generated
\ No newline at end of file
+generated
+.env
```
yoshiwarab, line 3: Are we ever running these tests with non-local env vars? If not maybe we want a committed test env file?

tests/e2e/runner/.gitignore
```diff
@@ -0,0 +1 @@
+.env
```
yoshiwarab, line 1: Are we ever running these tests with non-local env vars? If not maybe we want a committed test env file?

tests/e2e/runner/__tests__/base.test.ts
```diff
@@ -1,78 +1,42 @@
...
+  createSubgraphClient,
+  waitUntilSubgraphIsSynced,
+  getSubgraphConfig,
+} from "./utils/helpers";
 
-const contracts = JSON.parse(
-  fs.readFileSync("/usr/runner/shared/test-config.json", "utf-8")
-) as SubgraphConfig;
+// waiting for subgraph to sync can take longer than the default 5s timeout
+jest.setTimeout(30 * 1000);
```
yoshiwarab, line 14: Kinda surprised by this since honestly, since everything's on a local network would expect things to sync very quickly probably worth looking into in the future.

tests/e2e/runner/__tests__/graphql/get-contracts.graphql
```diff
@@ -1,5 +1,15 @@
+fragment contractDetails on Contract {
```
yoshiwarab, line 1: Nit: Uppercase `ContractDetails`

tests/e2e/runner/__tests__/graphql/get-core-registries.graphql
```diff
@@ -0,0 +1,15 @@
+fragment coreRegistryDetails on CoreRegistry {
```
yoshiwarab, line 1: Nit: uppercase CoreRegistryDetails

tests/e2e/runner/__tests__/graphql/get-minter-filters.graphql
```diff
@@ -0,0 +1,34 @@
+fragment minterFilterDetails on MinterFilter {
```
yoshiwarab, line 1: Nit: uppercase

tests/e2e/runner/__tests__/graphql/get-minters.graphql
```diff
@@ -0,0 +1,22 @@
+fragment minterDetails on Minter {
```
yoshiwarab, line 1: Nit: uppercase

tests/e2e/runner/__tests__/graphql/get-projects.graphql
```diff
@@ -0,0 +1,40 @@
+fragment projectDetails on Project {
```
yoshiwarab, line 1: Nit: uppercase

tests/e2e/seed/scripts/deploy.ts
```diff
@@ -24,43 +28,65 @@ import { JsonRpcProvider } from "@ethersproject/providers";
...
+  /////////////////////////////////////////////////////////////////////////////
   const accounts = JSON.parse(
     fs.readFileSync("./shared/accounts.json", "utf8")
   );
   const deployer = ethers.Wallet.fromMnemonic(accounts.mnemonic).connect(
     new JsonRpcProvider("http://hardhat:8545")
   );
+  const artist = ethers.Wallet.fromMnemonic(
+    accounts.mnemonic,
+    "m/44'/60'/1'/0/1" // derivation path index 1
```
yoshiwarab, line 64: Just curious, what is this exactly?

yoshiwarab (approved):
🙌 🙌 🙌

tests/e2e/runner/__tests__/base.test.ts
ryley-o: Ah, I added this at one point, but I think I realized later the delay was being caused by something else that I fixed. I've removed these timeout overrides and should be fine without them!

tests/e2e/runner/__tests__/graphql/get-minters.graphql
ryley-o: updated

tests/e2e/runner/__tests__/graphql/get-projects.graphql
ryley-o: updated

tests/e2e/runner/__tests__/graphql/get-minter-filters.graphql
ryley-o: updated

tests/e2e/runner/__tests__/graphql/get-core-registries.graphql
ryley-o: updated

tests/e2e/runner/__tests__/graphql/get-contracts.graphql
ryley-o: updated

src/mapping-v3-core.ts
```diff
@@ -31,10 +33,12 @@ import {
 } from "../generated/AdminACLV0/IAdminACLV0";
 
 import { MinterFilterV1 } from "../generated/MinterFilterV1/MinterFilterV1";
+import { loadOrCreateMinterFilter as legacyLoadOrCreateLegacyMinterFilter } from "./legacy-minter-filter-mapping";
```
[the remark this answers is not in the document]
ryley-o: oops! updated to be `legacyLoadOrCreatMinterFilter` (to align with naming pattern used in `legacyPopulateAllExistingMinterConfigurations`)

src/mapping-v3-core.ts
```diff
@@ -1204,30 +1235,132 @@ function populateAllExistingMinterConfigurations<T>(
...
+  }
+}
+
+/**
+ * Populate all project minter configurations from a given shared MinterFilter
+ * (e.g. IMinterFilterV1)
+ * @dev for legacy minter filters, use function
+ * legacyPopulateAllExistingMinterConfigurations
+ */
+function PopulateAllExistingMinterConfigurations<T>(
```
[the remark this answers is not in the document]
ryley-o: updated

tests/e2e/runner/.dockerignore
ryley-o: I can't think of a reason why we would be running these tests with non-local env vars (since they are so tightly coupled with the seed subgraph config). Part of this was just to update these to be consistent with our .gitignore, so we don't have different behavior in docker on different machines.

I was playing around with running codegen locally, which requires a populated environment var, so @yoshiwarab do you agree that it seems preferable to commit a .env file instead of defining the value in the docker-compose yaml?

yoshiwarab: Ah forgot they were already in docker compose,   let's leave this then!

ryley-o: okay - will leave as-is 👍

tests/e2e/runner/.gitignore
ryley-o: (see resolved comment in https://github.com/ArtBlocks/artblocks-subgraph/pull/223#discussion_r1251182993)

src/shared-minter-filter-mapping.ts
ryley-o: I've updated this section of code to be more readable, using the variable name `currentMinterFilter` 👍

tests/e2e/seed/scripts/deploy.ts
ryley-o: It is the bip-44 derivation path for account 1.
I slightly messed this up and should have used `address_index=0`, so I've updated and added a little more context:
```
"m/44'/60'/1'/0/0" // bip-44 derivation path Ethereum account 1`
```
The intent here is just to get the "next" account from the seed phrase that was stored during the deployment process, and assign it to artist.
refs:
- https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#path-levels
- https://docs.ethers.org/v5/api/signer/#Wallet.fromMnemonic

tests/e2e/runner/__tests__/base.test.ts
ryley-o: @yoshiwarab - I removed this, then actually did encounter a failure in `minter-filter-v2.test.ts` on one of the tests that performs a bunch of operations, waits for sync, queries subgraph, performs more operations, waits for sync, queries, and so on...
So I did bump up that file's timeout to 15s. Happy to chat more on this if you want!

src/helpers.ts
```diff
@@ -184,9 +304,23 @@ export function loadOrCreateMinter(
...
+    log.warning(
+      "[WARN] Minter at {} does not have a valid minter filter address",
+      [minterAddress.toHexString()]
+    );
+    const dummyMinterFilter = loadOrCreateSharedMinterFilter(
+      Address.zero(),
+      timestamp
+    );
+    minter.minterFilter = dummyMinterFilter.id;
+  }
```
jakerockland, line 323: Very much a nit, but I find reducing boolean logic where possible is a bit more readable if we need to handle both the if and the else.
```suggestion
  if (minterFilterResult.reverted) {
    // if minterFilterAddress() reverts, then the minter is not as expected and
    // we log warning, and assign to dummy MinterFilter entity at zero address
    log.warning(
      "[WARN] Minter at {} does not have a valid minter filter address",
      [minterAddress.toHexString()]
    );
    const dummyMinterFilter = loadOrCreateSharedMinterFilter(
      Address.zero(),
      timestamp
    );
    minter.minterFilter = dummyMinterFilter.id;
  } else {
    minter.minterFilter = minterFilterResult.value.toHexString();
  }
```

src/mapping-v3-core.ts
```diff
@@ -1204,30 +1235,132 @@ function populateAllExistingMinterConfigurations<T>(
...
+  timestamp: BigInt
+): void {
+  if (
+    !(
+      contract instanceof GenArt721CoreV3 ||
+      contract instanceof GenArt721CoreV3_Engine ||
+      contract instanceof GenArt721CoreV3_Engine_Flex
+    )
+  ) {
+    return;
```
jakerockland, line 1295: Nit: do we want to log a warning here before we return?

jakerockland (approved):

src/shared-minter-filter-mapping.ts
```diff
@@ -0,0 +1,344 @@
...
+import {
+  loadOrCreateSharedMinterFilter,
+  loadOrCreateMinter,
+  loadOrCreateCoreRegistry,
+  generateMinterFilterContractAllowlistId,
+  generateContractSpecificId,
+  loadOrCreateAndSetProjectMinterConfiguration
+} from "./helpers";
+
+export function handleDeployed(event: Deployed): void {
```
jakerockland, line 32:
```suggestion
/**
 * Handle the `Deployed` event. Creates a new MinterFilter entity in the store
 * to keep track of whether a MinterFilter is in our subgraph configuration.
 *
 * @param event - The `Deployed` event emitted by the contract
 */
export function handleDeployed(event: Deployed): void {
```

src/shared-minter-filter-mapping.ts
```diff
@@ -0,0 +1,344 @@
...
+  // we simply create a new MinterFilter entity to ensure that it is in the
+  // store. This enables us to determine if a MinterFilter is in our subgraph
+  // configuration by checking if it is in the store.
+  loadOrCreateSharedMinterFilter(event.address, event.block.timestamp);
+}
+
+export function handleMinterApprovedGlobally(
```
jakerockland, line 39:
```suggestion
/**
 * Handle the `MinterApprovedGlobally` event. Updates the globally allowlisted state
 * of a minter and add the minter to the list of globally allowlisted minters.
 *
 * @param event - The `MinterApprovedGlobally` event emitted by the contract
 */
export function handleMinterApprovedGlobally(
```

src/shared-minter-filter-mapping.ts
```diff
@@ -0,0 +1,344 @@
...
+  if (!minterFilter.minterGlobalAllowlist.includes(minter.id)) {
+    minterFilter.minterGlobalAllowlist = minterFilter.minterGlobalAllowlist.concat(
+      [minter.id]
+    );
+    minterFilter.updatedAt = event.block.timestamp;
+    minterFilter.save();
+  }
+}
+
+export function handleMinterRevokedGlobally(
```
jakerockland, line 75:
```suggestion
/**
 * Handle the `MinterRevokedGlobally` event. Updates the globally allowlisted state
 * of a minter and remove the minter from the list of globally allowlisted minters.
 *
 * @param event - The `MinterRevokedGlobally` event emitted by the contract
 */
export function handleMinterRevokedGlobally(
```

src/shared-minter-filter-mapping.ts
```diff
@@ -0,0 +1,344 @@
...
+    if (minterFilter.minterGlobalAllowlist[i] != minter.id) {
+      newMinterGlobalAllowlist.push(minterFilter.minterGlobalAllowlist[i]);
+    }
+  }
+  minterFilter.minterGlobalAllowlist = newMinterGlobalAllowlist;
+  minterFilter.updatedAt = event.block.timestamp;
+  minterFilter.save();
+}
+
+export function handleMinterApprovedForContract(
```
jakerockland, line 111:
```suggestion
/**
 * Handle the `MinterApprovedForContract` event. Updates the contract-specific allowlisted
 * state of a minter and add the minter to the list of contract-specific allowlisted minters.
 *
 * @param event - The `MinterApprovedForContract` event emitted by the contract
 */
export function handleMinterApprovedForContract(
```

src/shared-minter-filter-mapping.ts
```diff
@@ -0,0 +1,344 @@
...
+  if (!contractAllowlist.minterContractAllowlist.includes(minter.id)) {
+    contractAllowlist.minterContractAllowlist = contractAllowlist.minterContractAllowlist.concat(
+      [minter.id]
+    );
+    contractAllowlist.updatedAt = event.block.timestamp;
+    contractAllowlist.save();
+  }
+}
+
+export function handleMinterRevokedForContract(
```
jakerockland, line 149:
```suggestion
/**
 * Handle the `MinterRevokedForContract` event. Updates the contract-specific allowlisted
 * state of a minter and removes the minter from the list of contract-specific allowlisted minters.
 * If the contract allowlist is empty, the entity is deleted.
 *
 * @param event - The `MinterRevokedForContract` event emitted by the contract
 */
export function handleMinterRevokedForContract(
```

src/shared-minter-filter-mapping.ts
```diff
@@ -0,0 +1,344 @@
...
+  contractAllowlist.updatedAt = event.block.timestamp;
+  contractAllowlist.save();
+
+  // if the contract allowlist is empty, delete the entity
+  if (contractAllowlist.minterContractAllowlist.length == 0) {
+    store.remove("MinterFilterContractAllowlist", contractAllowlist.id);
+  }
+}
+
+export function handleProjectMinterRegistered(
```
jakerockland, line 194:
```suggestion
/**
 * Handle the `ProjectMinterRegistered` event. Updates the minter configuration of a project.
 * If the project does not exist or its minter filter does not match the minter filter that 
 * emitted the event, warnings are logged and the function returns early.
 *
 * @param event - The `ProjectMinterRegistered` event emitted by the contract
 */
export function handleProjectMinterRegistered(
```

src/shared-minter-filter-mapping.ts
```diff
@@ -0,0 +1,344 @@
...
+  // Create project minter configuration if needed, assign it to the project,
+  // and save
+  loadOrCreateAndSetProjectMinterConfiguration(
+    project,
+    minter,
+    event.block.timestamp
+  );
+}
+
+export function handleProjectMinterRemoved(event: ProjectMinterRemoved): void {
```
jakerockland, line 254:
```suggestion
/**
 * Handle the `ProjectMinterRemoved` event. Clears the minter configuration of a project.
 * If the project does not exist or its minter filter does not match the minter filter that
 * emitted the event, the function returns early.
 *
 * @param event - The `ProjectMinterRemoved` event emitted by the contract
 */
export function handleProjectMinterRemoved(event: ProjectMinterRemoved): void {
```

src/shared-minter-filter-mapping.ts
```diff
@@ -0,0 +1,344 @@
...
+    return;
+  }
+
+  // clear the project's minter configuration
+  project.minterConfiguration = null;
+  project.updatedAt = event.block.timestamp;
+  project.save();
+}
+
+export function handleCoreRegistryUpdated(event: CoreRegistryUpdated): void {
```
jakerockland, line 288:
```suggestion
/**
 * Handle the `CoreRegistryUpdated` event. Updates the core registry of a minter filter.
 *
 * @param event - The `CoreRegistryUpdated` event emitted by the contract
 */
export function handleCoreRegistryUpdated(event: CoreRegistryUpdated): void {
```

jakerockland (approved):

src/helpers.ts
ryley-o: yeah this makes sense to me, thanks!

src/mapping-v3-core.ts
ryley-o: We do this primarily because AS doesn't support union types, so we are stuck using generic type for contract. I would generally consider this as ~boilerplate, and if we were to ever call this function with a different type than the three items listed here, we should definitely be picking up in our test suite.

I think this pattern is followed many places in our handler codebase, so @jakerockland @yoshiwarab - what are your thoughts on if we should log a warning when type checking?

jakerockland: I see I see – my general inclination would be that it would be good practice to log warning when typechecking, but doing so as a later cleanup type item down the road seems perfectly acceptable. 

If this is a common boilerplate issue, could be worth writing a light typechecker utility that abstracts this check+warning-log behavior.

ryley-o: This all sounds good to me - I've added a polish task for this! https://app.asana.com/0/1203634050572702/1205096158411073/f

PR #225, opened 2023-05-25
Supplemental ABIs for unreleased contracts

ryley-o:
## Description of the change

Change legacy to supplemental abis to enable fast develop when e2e testing.

This adds a recommended process for devs to include unreleased ABIs in the `abis-supplemental directory`, which then enables e2e testing to work.

Previously, we would often locally change the `@artblocks/artblocks-contracts` dependency to point to a local repo when capturing unreleased changes, but that process is not (easily) compatible with e2e testing, where we want to install the package in a docker container.(https://docs.artblocks.io/creator-docs/art-blocks-api/queries/) sections).


jakerockland:
@ryley-o want to double-click on if this approach is preferred to just pulling in multiple versions of the contracts NPM package (old and newer versions, as needed)?

ryley-o:
>@ryley-o want to double-click on if this approach is preferred to just pulling in multiple versions of the contracts NPM package (old and newer versions, as needed)?

This approach specifically handles the case where we have (in development) ABIs that are not yet published in our npm package, but are required for our subgraph development. We used to just locally edit `package.json` to a local clone of the artblocks-contracts repo on the branch with the new ABIs, but that isn't compatible with running e2e tests in Docker (without copying the entire `node_modules` directory, which doesn't seem ideal).

This was where Ben Y and I landed after some discussion about different options.

re: old ABIs that are no longer included in the npm package - the legacy ABIs in the supplemental directory were never published in even our initial NPM package due to how old they were, but in general, agree with the take that we could use multiple versions of the NPM package if we want to remove ABIs from the NPM package in the future 👍

jakerockland:
> > @ryley-o want to double-click on if this approach is preferred to just pulling in multiple versions of the contracts NPM package (old and newer versions, as needed)?
> 
> This approach specifically handles the case where we have (in development) ABIs that are not yet published in our npm package, but are required for our subgraph development. We used to just locally edit `package.json` to point to a local clone of the artblocks-contracts repo on the branch with the new ABIs, but that isn't compatible with running e2e tests in Docker (without copying the entire `node_modules` directory, which doesn't seem ideal).
> 
> This solution was where Ben Y and I landed after some discussion about different options.
> 
> re: old ABIs that are no longer included in the npm package - the legacy ABIs in the supplemental directory were never published in even our initial NPM package due to how old they were, but in general, agree with the take that we could use multiple versions of the NPM package if we want to remove ABIs from the NPM package in the future 👍

SGTM ser – thank you very much for the elaboration here. 💜

ryley-o:
@yoshiwarab - gentle ping on this relatively simple one 🙏

yoshiwarab (approved):

PR #229, opened 2023-06-15
Phase II.C: Handlers for ISharedMinterV0

ryley-o:
## Description of the change

Update subgraph with handlers for new `ISharedMinterV0` interface events.

e2e tests are also written for the events, with the caveat that the event `ProjectCurrencyInfoUpdated` is not able to be tested. This is because only an ERC20 minter emits that event, and the ERC20 minter is still in development.


src/i-shared-minter-mapping.ts
```diff
@@ -0,0 +1,163 @@
...
+  );
+
+  projectMinterConfig.currencyAddress = event.params._currencyAddress;
+  projectMinterConfig.currencySymbol = event.params._currencySymbol;
+  projectMinterConfig.save();
+
+  // update project's updatedAt timestamp to induce a sync
+  // @dev this may induce a sync if the project's minter is not this minter,
+  // but good to consider the project as "updated", as it could switch to this
+  // minter in the future
```
yoshiwarab, line 93: Wouldn't we sync the project with the updated minter config once the minter is assigned to the project anyways? Also okay with leaving this here regardless so that our db row for the minter config is always up to date. We should probably consider syncing minter configs separately in the future though assuming we want to sync all changes.

src/i-shared-minter-mapping.ts
```diff
@@ -0,0 +1,163 @@
...
+// HELPER FUNCTIONS start here
+///////////////////////////////////////////////////////////////////////////////
+/**
+ * Helper function to attempt to load a project from the store. If the project
+ * does not exist, returns null.
+ * @param coreContractAddress core contract address of the project
+ * @param projectNumber project number of the project (BigInt)
+ * @returns The Project entity from the store if it exists, otherwise null
+ */
+function tryLoadProject(
```
yoshiwarab, line 150: Nit: `tryLoadProject` feels a bit misleading as a name here because of the association with `try` being something that would catch an error that we could handle. I think just `loadProjectByCoreAddressAndProjectNumber` or something would be fine since the behavior of returning null is inline with the normal entity load function.

tests/e2e/runner/__tests__/graphql/get-project-minter-configurations.graphql
```diff
@@ -0,0 +1,28 @@
+fragment projectMinterConfigurationDetails on ProjectMinterConfiguration {
```
yoshiwarab, line 1: Nit: uppercase

yoshiwarab (approved):

tests/e2e/runner/__tests__/graphql/get-project-minter-configurations.graphql
ryley-o: updated 🙏

src/i-shared-minter-mapping.ts
ryley-o: agreed, updated to suggested name in [67668ca](https://github.com/ArtBlocks/artblocks-subgraph/pull/229/commits/67668ca36c262bc1fc42b341b054d42fbb0d86f7)

src/i-shared-minter-mapping.ts
ryley-o: Thanks for this comment. It is true that we only induce syncs of the active projectMinterConfig, so I think we should only update project `updatedAt` to induce that sync if the projectMinterConfig is the current project minter configuration.
Updated logic in commit [b89ad8d](https://github.com/ArtBlocks/artblocks-subgraph/pull/229/commits/b89ad8d1e027d00d38cc16b2fe3c59679b65ab3a)

src/i-shared-minter-mapping.ts
```diff
@@ -0,0 +1,164 @@
...
+  loadOrCreateMinter,
+  generateContractSpecificId,
+  loadOrCreateProjectMinterConfiguration
+} from "./helpers";
+
+///////////////////////////////////////////////////////////////////////////////
+// EVENT HANDLERS start here
+///////////////////////////////////////////////////////////////////////////////
+
+export function handlePricePerTokenInWeiUpdated(
```
jakerockland, line 20:
```suggestion
/**
 * Handles the update of price per token in wei. Attempts to load associated project and 
 * its minter configuration, then updates base price in the configuration.
 * @param event The event carrying new price per token in wei
 */
export function handlePricePerTokenInWeiUpdated(
```

src/i-shared-minter-mapping.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  // only induce sync via updating project's updatedAt if the
+  // projectMinterConfig is the active minter configuration for the project
+  if (project.minterConfiguration == projectMinterConfig.id) {
+    project.updatedAt = event.block.timestamp;
+    project.save();
+  }
+}
+
+export function handleProjectCurrencyInfoUpdated(
```
jakerockland, line 59:
```suggestion
/**
 * Handles the update of a project's currency information. Attempts to load associated 
 * project and its minter configuration, then updates currency address and symbol.
 * @param event The event carrying updated currency information
 */
export function handleProjectCurrencyInfoUpdated(
```

src/i-shared-minter-mapping.ts
```diff
@@ -0,0 +1,164 @@
...
+  projectMinterConfig.basePrice = event.params._pricePerTokenInWei;
+  projectMinterConfig.priceIsConfigured = true;
+  projectMinterConfig.save();
+
+  // only induce sync via updating project's updatedAt if the
+  // projectMinterConfig is the active minter configuration for the project
+  if (project.minterConfiguration == projectMinterConfig.id) {
+    project.updatedAt = event.block.timestamp;
+    project.save();
+  }
```
jakerockland, line 56: Nit: we replicate this in multiple places, may be worth splitting into a smol util if we expect that this may drift otherwise.

src/i-shared-minter-mapping.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+export function handleProjectMaxInvocationsLimitUpdated(
```
jakerockland, line 98:
```suggestion
/**
 * Handles the update of a project's max invocations limit. Attempts to load associated 
 * project and its minter configuration, then updates max invocations in the configuration.
 * @param event The event carrying updated max invocations limit
 */
export function handleProjectMaxInvocationsLimitUpdated(
```

tests/e2e/runner/__tests__/minter-suite-v2/i-shared-minter-v0.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  // ERC20 minter is not yet developed at the time of writing this test
+  // describe("ProjectCurrencyInfoUpdated", () => {
+  //   afterEach(async () => {
+  //     // TODO
+  //   });
+
+  //   test("Currency is updated and configured", async () => {
+  //     // TODO - need ERC20 minter to test this
+  //   });
+  // });
```
jakerockland, line 132: Can we please add an associated Asana ticket linked here for this TODO?

jakerockland (approved):

src/i-shared-minter-mapping.ts
ryley-o: good call, we have this many places 😅 (including downstream PRs). I'll update and propagate downstream 👍

tests/e2e/runner/__tests__/minter-suite-v2/i-shared-minter-v0.test.ts
ryley-o: yeah it is added in #235 (which is tracked in Asana here: https://app.asana.com/0/0/1205078004803554/f)
- @jakerockland would it be best to commit the asana link to our codebase in these situations?

src/i-shared-minter-mapping.ts
ryley-o: added in [8551d76](https://github.com/ArtBlocks/artblocks-subgraph/pull/229/commits/8551d76335f6d613735ced5a15bc0a820f913edf), and will propagate to other PRs now

tests/e2e/runner/__tests__/minter-suite-v2/i-shared-minter-v0.test.ts
ryley-o: oh, I see what you were saying - yep we have it in asana (and already in a PR up for review). resolving 😄

PR #232, opened 2023-06-23
Phase II.E: Handlers for ISharedMinterHolderV0

ryley-o:
## Description of the change

Add handlers for IFilteredMinterHolderV0.

## Tests
- e2e tests are added for all events in `IFilteredMinterHolderV0`
- e2e tests are NOT added for events already tested in the `ISharedMinterV0` interface, since the logic in those handlers is already tested with other minters


src/i-shared-minter-mapping.ts
```diff
@@ -349,7 +349,7 @@ function tryLoadProject(
   return Project.load(fullProjectId);
 }
 
-function loadOrCreateMinterProjectAndConfigIfProject(
+export function loadOrCreateMinterProjectAndConfigIfProject(
```
ryley-o, line 352: this could go in a common helpers file, but kind of liked keeping it in this file, as it is intended to be used for the new shared minters 🤷

yoshiwarab (approved):

jakerockland (approved):

PR #238, opened 2023-07-25
Update to remove `AtIndex` from dependency registry gets

jakerockland:
## Description of the change

Minor adjustment to be consistent with consistency adjustments made in: https://github.com/ArtBlocks/artblocks-contracts/pull/895


jakerockland:
@ryley-o @yoshiwarab am I right in my understanding that I need to cut a new version of the contracts ABI package before the ABI generation here will work as intended (and unblock the breaking tests)?

jakerockland:
cc @sterlingcrispin for vis on how these contracts + subgraph PRs inter-relate

ryley-o:
We already ended up implementing this in the minter suite V2 branch, which was merged to main today. It was required when we updated to use the new contracts package 👍

PR #240, opened 2023-07-31
Index minter filter type

yoshiwarab:
## Description of the change
Updates subgraph to index minter filter type. Note: this will need to be added within the new minter suite architecture handlers as well, cc: @ryley-o 

### DIFF
--- a/src/legacy-minter-filter-mapping.ts
+++ b/src/legacy-minter-filter-mapping.ts
@@ -6,6 +6,8 @@ import {
   IsCanonicalMinterFilter
 } from "../generated/MinterFilterV0/MinterFilterV0";
 
+import { MinterFilterV1 } from "../generated/MinterFilterV1/MinterFilterV1";
+
 import {
   IMinterFilterV0,
   Deployed,
@@ -63,6 +65,7 @@ export function handleIsCanonicalMinterFilter(
     );
     minterFilter.coreRegistry = coreRegistry.id;
     minterFilter.minterGlobalAllowlist = [];
+    minterFilter.type = "MinterFilterV0";
     minterFilter.updatedAt = event.block.timestamp;
     minterFilter.save();
   }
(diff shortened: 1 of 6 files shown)


src/legacy-minter-filter-mapping.ts
```diff
@@ -337,6 +340,18 @@ export function loadOrCreateMinterFilter(
...
+
+  // IMinterFilterV0 has no minterFilterType() function, however MinterFilterV1
+  // which implements IMinterFilterV0 does have a minterFilterType() function.
+  // We can use this to determine if the minter filter is a V0 or V1 minter filter.
+  const potentialV1MinterFilter = MinterFilterV1.bind(minterFilterAddress);
+  let minterTypeResult = potentialV1MinterFilter.try_minterFilterType();
+  if (!minterTypeResult.reverted) {
+    minterFilter.type = minterTypeResult.value;
+  } else {
+    minterFilter.type = "MinterFilterV0";
```
ryley-o, line 352: This can actually happen on a MinterFilterV1, specifically for ones deployed before 2023, including our Flagship MinterFilterV1 here 😓: https://etherscan.io/address/0x092B8F64e713d66b38522978BCf4649db14b931E#readContract  

Even more frustrating, MinterFilterV0 and V1 have the same read functions!
There might be hope though - if we can read the revert reason, `alertAsCanonicalMinterFilter` only exists on V0 and would revert with "Only Core whitelisted" , whereas on V1 it would revert because the function wouldn't be found..

src/legacy-minter-filter-mapping.ts
```diff
@@ -337,6 +340,28 @@ export function loadOrCreateMinterFilter(
...
+  // IMinterFilterV0 contracts may or may not have a minterFilterTypeFunction
+  // so we derive this based on the associated core contract address instead.
+  // All pre GenArt721V3 contracts should have a V0 minter filter while all
+  // GenArt721V3 contracts and above should have a V1 minter filter or higher.
+  // If the minter filter version is greater than V1 we can assume the entity
+  // creation will be handled by another event handler.
+  const coreContract = Contract.load(coreContractAddress.toHexString());
+  // We don't expect this actually happens, but we'll log a warning if it does.
+  if (!coreContract) {
+    log.warning(
```
yoshiwarab, line 353: I've chosen not to early return here so we don't need to update anywhere that might call this function. We shouldn't ever really hit this and I believe the go wrong case is pretty harmless.

ryley-o (approved):
LGTM - I'll follow up on the `minter-suite-v2-phase-II` branch 🙏

jakerockland (approved):

PR #243, opened 2023-08-07
handle unknown minter filter type

ryley-o:
## Description of the change

Handle unknown minter filter type.

This is intended to add robustness to our subgraph such that it won't crash when the follow situation is encountered:
- Engine partner wants to use their own minter filter
- Engine partner does implement the function `minterFilterType`
- The returned value from `minterFilterType` is not in the minter filter types enum

>Note: The current behavior if a partner were to NOT implement the `minterFilterType` function would be to cast the MinterFilter as a `MinterFilterV1` if its core contract starts with `GenArt721CoreV3`, or `MinterFilterV0` otherwise, since some of those deploiyed minterfilters did not implement the `minterFilterType` function (see functions `getIsLegacyMinterFilter` and `legacy-minter-filter-mapping.ts:loadOrCreateMinterFilter`)

This implementation adds a new `UNKNOWN` value to the minter filter types enum. Another option would be to simply record the unrecognized type as a `MinterFilterV2` and log a warning. Would appreciate feedback in this area, as it could impact downstream complexity (due to potentially having to handle the `UNKNOWN` type in downstream database, SDKs, etc.).

---
  - https://app.asana.com/0/0/1205206309483255


jakerockland (approved):

ryley-o:
ah, I missed this one yesterday - gentle ping @lyaunzbe on this one, as it would be good to get in before we merge all Phase II updates into main 🙏

lyaunzbe (approved):

PR #244, opened 2023-08-07
Phase II.I: Shared DA Minters w/Settlement

ryley-o:
## Description of the change

Add handlers for shared DA minters w/Settlement.

The settlement minters have a new interface, `ISharedMinterDAExpSettlementV0`, which emits a few events that must be handled.

## Tests
- E2E tests are added for the new event handler
- No new tests are required for the two generic events utilized by the new minter, because they already are tested in `i-shared-minter-v0-generic-handlers.test.ts` 🥳 

---

Asana task: https://app.asana.com/0/0/1205175206458362/f


jakerockland (approved):

ryley-o:
ah, these tests are failing due to dependency issues that are fixed in #246. I think it is fine to merge, as they were passing prior to some non-changing recent merges, and we have a fix in the next PR, and this is only being merged into a feature branch.

lyaunzbe (approved):

PR #247, opened 2023-09-07
Bug fix current settled price type

ryley-o:
## Description of the change

Update generic minter event handlers to convert BigInt prices to strings. (also fix bug where num settleable invocations is converted from BigInt to string)

Currently, we always convert prices stored in `extraMinterDetails` json fields to strings. This ensures that we do not encounter numeric overflow when loading the json field's data in our frontend, because javascript's max number value is less than common ETH prices (1e18+).

For the new minter suite, for the DA Settlement V3 minter, a generic event is emitted that updates latest purchase price in extra minter details. This is the first instance of a price being updated via generic event in our subgraph. It causes a bug where the price is stored as a number instead of a string, because the generic event indeed communicates the price as a number. This throws a wrench in our downstream systems, including Hasura's calculation of available excess settlement funds.

The fix proposed in this PR is to convert any generic number values whose key contains substring "price" or "Price" to a string. This fixes the issue described above, because the key is named `"currentSettledPrice"`.

In addition to to `currentSettledPrice`, `numSettleableInvocations` was incorrectly being stored as a string by our handler, which is inconsistent with behavior on legacy settlement minters. This is also fixed as part of this PR.

## Alternatives considered

A different heuristic could be used to determine if a generic event's value should be converted to a string, and a few were considered:
- Value-based: BigInt values greater than a certain value could be converted to a string. Issue with this is that low prices are possible, and having uncertainty downstream as to what a key's value type will be is not a good state.
- List of keys to be converted: Instead of looking for the phrase "price" in key name, we could check against an array of keys and convert if the key is in the array. This seems like a valid solution, but seemed a little more complex than the selected implementation. Open to feedback here for sure though!

## Test

- e2e test logic was added to expose this issue with a test failure prior to implementing the fix
- the updated handler logic allows the e2e tests to pass ✅ 

## Other

Supplemental ABIs are added for e2e testing, because the latest contracts npm package release has a V3 minter that doesn't emit a proper event for `currentSettledPrice` in this exact situation 🤣

### DIFF
--- a/tests/e2e/runner/__tests__/minter-suite-v2/i-filtered-shared-da-exp-settlement.test.ts
+++ b/tests/e2e/runner/__tests__/minter-suite-v2/i-filtered-shared-da-exp-settlement.test.ts
@@ -22,8 +22,10 @@ import { Logger } from "@ethersproject/logger";
 Logger.setLogLevel(Logger.levels.ERROR);
 
 // waiting for subgraph to sync can take longer than the default 5s timeout
-// @dev specifically for this test, we need to wait for auction to complete, so need extra time
-jest.setTimeout(75 * 1000);
+// @dev specifically for this test, we need to wait for auction to complete, which is half of 45-second min half life
+// length = 22.5 seconds, plus 6 buffer seconds for the auction to start ~=30 seconds,
+// times 2 margin since this is just a timeout = 60 seconds timeout
+jest.setTimeout(60 * 1000);
 
 const config = getSubgraphConfig();
 
--- a/tests/e2e/runner/__tests__/minter-suite-v2/i-filtered-shared-da-exp-settlement.test.ts
+++ b/tests/e2e/runner/__tests__/minter-suite-v2/i-filtered-shared-da-exp-settlement.test.ts
@@ -22,7 +22,8 @@ import { Logger } from "@ethersproject/logger";
 Logger.setLogLevel(Logger.levels.ERROR);
 
 // waiting for subgraph to sync can take longer than the default 5s timeout
-jest.setTimeout(30 * 1000);
+// @dev specifically for this test, we need to wait for auction to complete, so need extra time
+jest.setTimeout(75 * 1000);
 
 const config = getSubgraphConfig();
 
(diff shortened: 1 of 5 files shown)


lyaunzbe (approved):

tests/e2e/runner/__tests__/minter-suite-v2/i-filtered-shared-da-exp-settlement.test.ts
```diff
@@ -22,7 +22,8 @@ import { Logger } from "@ethersproject/logger";
 Logger.setLogLevel(Logger.levels.ERROR);
 
 // waiting for subgraph to sync can take longer than the default 5s timeout
-jest.setTimeout(30 * 1000);
+// @dev specifically for this test, we need to wait for auction to complete, so need extra time
+jest.setTimeout(75 * 1000);
```
aaronpenne, line 26: why 75? would like some context/clarity on the hardcoded number added to the comment

tests/e2e/runner/__tests__/minter-suite-v2/i-filtered-shared-da-exp-settlement.test.ts
```diff
@@ -155,6 +157,47 @@ describe("iSharedMinterDAExpSettlement event handling", () => {
...
+        );
+      const targetConfigId = `${minterDAExpSettlementV3Address.toLowerCase()}-${genArt721CoreAddress.toLowerCase()}-${currentProjectNumber.toString()}`;
+      const projectMinterConfigRes = await getProjectMinterConfigurationDetails(
+        client,
+        targetConfigId
+      );
+      const extraMinterDetails = JSON.parse(
+        projectMinterConfigRes.extraMinterDetails
+      );
+      // @dev important that this is a string due to potential for numeric overflow in js
```
aaronpenne, line 176: appreciate the clear comments

tests/e2e/runner/__tests__/minter-suite-v2/i-filtered-shared-da-exp-settlement.test.ts
ryley-o: good call - updated with some additional comments + reduced to 60s in [0458c18](https://github.com/ArtBlocks/artblocks-subgraph/pull/247/commits/0458c18608e0b7218d25fdf13206fc680632ef5c)

for additional context here, our e2e tests really do test things end-to-end, so in this case, we need to really run an auction, which is configured as 22.5 seconds. If we try and get too precise and shorten more, we would risk missing the target intermediate point in the auction. Time is a bit precise because we have a jest hardhat script talking to the locally running blockchain, and locally running subgraph, which is also talking to the local blockchain.

PR #262, opened 2023-11-15
Track purchase details on token mint

yoshiwarab:
## Description of the change
This PR adds a new PrimaryPurchaseDetails entity type that includes data about the minter address used to mint a token, as well as currency info. All core handlers have been updated to add this data. Pre-v3 contracts will not always have this populated as the necessary data is not always available from within the handler.

https://app.asana.com/0/1201568815538912/1205847400671374/f

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -349,6 +349,26 @@ type Token @entity {
 
   "Next available sale id"
   nextSaleId: BigInt!
+
+  primaryPurchaseDetails: PrimaryPurchaseDetails
+}
+
+type PrimaryPurchaseDetails @entity {
+  id: ID!
+
+  token: Token!
+
+  "Transaction hash of token purchase"
+  transactionHash: Bytes!
+
+  "Address of the minter used to mint token"
+  minterAddress: Bytes
+
+  "Address of currency used to purchase token"
+  currencyAddress: Bytes
+
+  "Symbol of currency used to purchase token"
+  currencySymbol: String
 }
 
 type MinterFilter @entity {
--- a/schema.graphql
+++ b/schema.graphql
@@ -354,6 +354,7 @@ type Token @entity {
 }
 
 type PrimaryPurchaseDetails @entity {
+  "Unique identifier made up of contract address and token id"
   id: ID!
 
   token: Token!
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -2,13 +2,11 @@ import {
   Address,
   BigInt,
   Bytes,
-  json,
   JSONValue,
-  JSONValueKind,
   TypedMap,
   store,
   log,
-  ethereum
+  ByteArray
 } from "@graphprotocol/graph-ts";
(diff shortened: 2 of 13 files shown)


yoshiwarab:
@ryley-o would you mind running the e2e tests when you get the chance, I've been unable to get the graph-node docker image to build on my new computer.

src/helpers.ts
```diff
@@ -657,3 +659,85 @@ export function snapshotStateAtSettlementRevenueWithdrawal(
...
+      purchaseDetails.minterAddress = changetype<Bytes>(
+        ByteArray.fromHexString(minterConfig.minter)
+      );
+      purchaseDetails.currencyAddress = minterConfig.currencyAddress;
+      purchaseDetails.currencySymbol = minterConfig.currencySymbol;
+
+      purchaseDetails.save();
+      return purchaseDetails.id;
+    } else {
+      log.warning("Minter configuration with id {} does not exists", [
```
ryley-o, line 717:
```suggestion
      log.warning("Minter configuration with id {} does not exist", [
```

schema.graphql
```diff
@@ -349,6 +349,26 @@ type Token @entity {
 
   "Next available sale id"
   nextSaleId: BigInt!
+
+  primaryPurchaseDetails: PrimaryPurchaseDetails
+}
+
+type PrimaryPurchaseDetails @entity {
+  id: ID!
```
ryley-o, line 357:
```suggestion
  "Unique identifier made up of contract address and token id"
  id: ID!
```

src/helpers.ts
```diff
@@ -657,3 +659,85 @@ export function snapshotStateAtSettlementRevenueWithdrawal(
...
+      ]);
+      return null;
+    }
+  } else {
+    // If no minter configuration is associated with the project, use the minter currently assigned to the contract
+    const contract = Contract.load(project.contract);
+    const to = changetype<Address>(event.transaction.to || Address.zero());
+
+    // If the transaction to address is a whitelisted minter, use it as the minter address
+    if (contract && contract.mintWhitelisted.includes(to)) {
```
ryley-o, line 728: In general, assuming that `tx.to` is the minter address is a good assumption, although it is not foolproof (e.g. bot minting from contract, etc.). I don't think it is worth adding much more complexity here, but perhaps before falling back to `Address.zero()`, you could check if the length of `contract.mintWhitelisted` == 1, in which case you could assume the single item in the array was the minter.

ryley-o (approved):
LGTM!

src/helpers.ts
yoshiwarab: > In general, assuming that tx.to is the minter address is a good assumption, although it is not bulletproof (e.g. bot minting from contract, etc.)

I think the check for `contract.mintWhitelisted.includes(to)` should confirm the assumption though no?

Also, added the check for only a single allowed minter, good call!

PR #254, opened 2023-10-18
License type dev

sterlingcrispin:
## Description of the change

adds License entity including events for LicenseTypeAdded and LicenseTextUpdated
the events and handlers associated with these
and test coverage

PR #265, opened 2023-12-05
Index license type on dependency registry

yoshiwarab:
## Description of the change

This PR is base off of https://github.com/ArtBlocks/artblocks-subgraph/pull/254 and includes updates to index license type on the Dependency entity type. Additionally it includes changes for compatibility with the latest updates to the DependencyRegistryV0 contract (see https://github.com/ArtBlocks/artblocks-contracts/pull/1215).


src/dependency-registry.ts
```diff
@@ -367,7 +415,7 @@ export function handleSupportedCoreContractRemoved(
   event: SupportedCoreContractRemoved
 ): void {
   const coreContract = Contract.load(
```
ryley-o, line 417: unrelated to this PR - probably worth a check that `coreContract.dependencyRegistry == event.address` before setting it to null? (I know we only expect one dependency registry, but who knows...)

src/dependency-registry.ts
```diff
@@ -379,42 +427,42 @@ export function handleSupportedCoreContractRemoved(
...
+    event.params.projectId
   );
 
   const project = Project.load(projectId);
   if (!project) {
     return;
   }
 
-  project.scriptTypeAndVersion = event.params._dependencyType.toString();
+  project.scriptTypeAndVersion = event.params.dependencyNameAndVersion.toString();
```
ryley-o, line 443: Also unrelated to this specific PR, but I don't think we should allow both core contract to set this field, and also the core registry. The state in the subgraph then depends on ordering, which might be pretty confusing? WDYT about a follow-on PR to do something like add a nullable `scriptTypeAndVersionOverride` field to project?

ryley-o (approved):
These changes make sense to me!
Added a couple notes on things that I think are worth touching on, but I think follow-on work since this PR is meant for naming updates.

src/dependency-registry.ts
yoshiwarab: Ah, yes definitely agree!

src/dependency-registry.ts
yoshiwarab: Agreed! https://app.asana.com/0/1201568815538912/1206099136189468/f

PR #270, opened 2023-12-19
Add split atomic indexing

ryley-o:
## Description of the change

Add indexing for split atomic factory and split atomic contracts.

Adds new entities, and includes new e2e tests.

Aligned with the contracts implementation in https://github.com/ArtBlocks/artblocks-contracts/pull/1292.

---
  - https://app.asana.com/0/0/1206266912756250


ryley-o:
12316d8 passes all e2e tests

lindsgil (approved):

yoshiwarab (approved):
LGTM! Wondering if we should publish a new @artblocks/contracts version so we don't need to add the abis-supplemental?

ryley-o:
>Wondering if we should publish a new @artblocks/contracts version so we don't need to add the abis-supplemental?

Good callout - due to prioritizing deployments on this, I'll do this in a follow-up PR. I think it will also let us remove the IDependencyRegistryV0.json from the abis-supplemental, so kind of fits nicely in a follow-on PR as well 😃 🙏

PR #271, opened 2023-12-20
Run E2E tests as part of CI

yoshiwarab:
## Description of the change

This PR updates our circle ci config and docker compose/docker files so that e2e scripts can be run as part of our CI.


tests/e2e/graph-node/Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM graph-node:latest
+FROM graphprotocol/graph-node:latest
```
yoshiwarab, line 1: @ryley-o you'll likely want to change this locally back to graph-node:latest so you can use your locally built graph-node image for M1

ryley-o: ah, yes.
WDYT about maybe just using an env var and defining is simple dockerfile in-line in `docker-compose.yml`?

ryley-o (approved):

tests/e2e/graph-node/Dockerfile
ryley-o: (then could just define appropriate graph-node image in `test:e2e:ci` and `test:e2e` scripts)

yoshiwarab: great idea! I added a separate script for m1 because I also pull the graphprotocol/graph-node image directly instead of rebuilding for m1 (m1 process didn't work for my m2 machine for some reason)

PR #279, opened 2024-01-24
Index SEA Bids

ryley-o:
## Description of the change

Index historical bids for SEA minter

Asana: https://app.asana.com/0/1201568815538912/1206375217836430/f

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -478,6 +478,33 @@ type ProjectMinterConfiguration @entity {
   maxInvocations: BigInt
 }
 
+type Bid @entity {
+  "Unique identifier made up of for SEA - minter-project-bidder-value-timestamp-token [TODO - do something different from RAM]"
+  id: ID!
+
+  "The associated project"
+  project: Project!
+
+  "The associated minter"
+  minter: Minter!
+
+  token: Token
+
+  "The associated account - TODO make this an Account"
+  bidder: Bytes!
+
+  "The value of the bid"
+  value: BigInt!
+
+  "TBD on a good description"
+  winningBid: Boolean!
+
+  "The timestamp of the bid"
+  timestamp: BigInt!
+
+  updatedAt: BigInt!
+}
+
 type Receipt @entity {
   "Unique identifier made up of {minter contract address}-{core contract address}-{project number}-{account address}"
   id: ID!
--- a/schema.graphql
+++ b/schema.graphql
@@ -488,6 +488,7 @@ type Bid @entity {
   "The associated minter"
   minter: Minter!
 
+  "The associated token"
   token: Token
 
   "The associated account"
--- a/schema.graphql
+++ b/schema.graphql
@@ -479,7 +479,7 @@ type ProjectMinterConfiguration @entity {
 }
 
 type Bid @entity {
-  "Unique identifier made up of for SEA - {minter contract address}-{bidder}-{timestamp}-{token} [TODO - do something different from RAM]"
+  "Unique identifier made up of for SEA - {minter contract address}-{bidder}-{value}-{token} [TODO - do something different from RAM]"
   id: ID!
 
   "The associated project"
--- a/schema.graphql
+++ b/schema.graphql
@@ -479,7 +479,7 @@ type ProjectMinterConfiguration @entity {
 }
 
 type Bid @entity {
-  "Unique identifier made up of for SEA - minter-project-bidder-value-timestamp-token [TODO - do something different from RAM]"
+  "Unique identifier made up of for SEA - {minter contract address}-{bidder}-{timestamp}-{token} [TODO - do something different from RAM]"
   id: ID!
 
   "The associated project"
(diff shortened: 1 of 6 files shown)


schema.graphql
```diff
@@ -478,6 +478,33 @@ type ProjectMinterConfiguration @entity {
   maxInvocations: BigInt
 }
 
+type Bid @entity {
+  "Unique identifier made up of for SEA - minter-project-bidder-value-timestamp-token [TODO - do something different from RAM]"
```
ryley-o, line 482: I think since we have token in this, we should remove project from this ID (since redundant).
Also, I'm thinking we should remove timestamp? Fine with indexing it as a field, but I don't think we need it for uniqueness, and just adds more info we need to be able to properly load a token, right?
thoughts?

lindsgil: Agreed on the project id! I'll remove that. I think we need to keep timestamp concatenated in the ID for the case of a user bidding -> being outbid -> bidding again with a higher value for the same token ID. That way we have a full historical record of all bids and not just their latest bid.

ryley-o: I think in any case, since the bidding -> being outbid -> bidding again can technically happen in a single block, timestamp still doesn't add any uniqueness. However, doesn't including value protect against that case?

src/sea-lib-mapping.ts
```diff
@@ -253,6 +280,63 @@ export function handleAuctionBid(event: AuctionBid): void {
...
+    // This is never expected to be possible with SEA minter
+    log.warning("Existing Bid with ID {} found for SEA Minter contract {}", [
+      bidId,
+      event.address.toString()
+    ]);
+  } else {
+    // New bid
+    bid = new Bid(bidId);
+    // Create new account entity if one for the bidder doesn't exist
+    const bidderAccount = new Account(event.params.bidder.toHexString());
```
asaltste, line 305: do we need to first try and load the account?

asaltste (approved):
LGTM thanks so much one question about new account creation but thats it

schema.graphql
lindsgil: Great point! - Right now it's `{minter}-{bidder}-{timestamp}-{token}` so I will update it to `{minter}-{bidder}-{value}-{token}`

src/sea-lib-mapping.ts
lindsgil: I think that if the account already exists it just overwrites the existing id with the same value, saving us from having to call load.. LMK if I'm missing anything here @ryley-o

ryley-o: yup! I think this is how we commonly deal with accounts since it only has an ID and derived fields 👍

ryley-o (commented):
LGTM!
(GH won't let me approve since I guess I made initial push of this)

schema.graphql
```diff
@@ -478,6 +478,33 @@ type ProjectMinterConfiguration @entity {
...
+  "The associated minter"
+  minter: Minter!
+
+  token: Token
+
+  "The associated account"
+  bidder: Account!
+
+  "The value of the bid"
+  value: BigInt!
```
ryley-o, line 497: @yoshiwarab can you remind me if having a price here will create problems downstream? I can't remember if we only needed a price-as-string in the json fields, or if we should also be using price here.

yoshiwarab: Yeah this should be fine and will just come through as a string when queried. As long as the corresponding DB field is numeric (supports very large numbers) we should be good.

src/sea-lib-mapping.ts
```diff
@@ -253,6 +276,63 @@ export function handleAuctionBid(event: AuctionBid): void {
...
+    event.params.tokenId.toString();
+
+  let bid = Bid.load(bidId);
+  if (bid) {
+    // This is never expected to be possible with SEA minter
+    log.warning("Existing Bid with ID {} found for SEA Minter contract {}", [
+      bidId,
+      event.address.toString()
+    ]);
+  } else {
```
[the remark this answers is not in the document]
lindsgil: Sounds good! Added

src/sea-lib-mapping.ts
```diff
@@ -253,6 +276,63 @@ export function handleAuctionBid(event: AuctionBid): void {
...
+
+  // Update Bids entity
+  const bidId =
+    event.address.toHexString() +
+    "-" +
+    event.params.bidder.toHexString() +
+    "-" +
+    event.params.bidAmount.toString() +
+    "-" +
+    event.params.tokenId.toString();
```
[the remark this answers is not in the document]
lindsgil: Good call! Done

src/sea-lib-mapping.ts
```diff
@@ -253,6 +276,63 @@ export function handleAuctionBid(event: AuctionBid): void {
...
+        previousHighestBidValueJSON.toString() +
+        "-" +
+        event.params.tokenId.toString();
+
+      const previousWinningBid = Bid.load(previousHighestBidId);
+      if (previousWinningBid) {
+        previousWinningBid.winningBid = false;
+        previousWinningBid.updatedAt = event.block.timestamp;
+        previousWinningBid.save();
+      }
```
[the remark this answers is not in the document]
lindsgil: Good call, it did feel a bit disjoint. Updated !

lindsgil (approved):

PR #280, opened 2024-02-16
RAM Minter Events + Tests

lindsgil:
## Description of the change

Added the Minter-Level Events:
- MinAuctionDurationSecondsUpdated
- MinterRefundGasLimitUpdated
- NumSlotsUpdated
- AuctionBufferTimeParamsUpdated

Project-Level Events:
- AuctionConfigUpdated
- NumTokensInAuctionUpdated
- AuctionTimestampEndUpdated

Bid/Purchase-Level Events:
- BidCreated
- BidRemoved
- BidToppedUp
- BidSettled
- BidMinted
- BidRefunded


abis-supplemental/MinterRAMV0.json
```diff
@@ -0,0 +1,1408 @@
+{
```
ryley-o, line 1: I don't think we should need any non-interface / non-library ABIs in the main subgraph - if something is missing from the interface, definitely feel free to add it to the contracts repo PR and update the interface or library here 👍

schema.graphql
```diff
@@ -495,11 +495,17 @@ type Bid @entity {
   bidder: Account!
 
   "The value of the bid"
-  value: BigInt!
+  value: BigInt
```
ryley-o, line 498: nit: I think still makes sense to calculate Bid Value for RAM minter - see solidity function in RAMLib and could duplicate that calc in the subgraph handler.

src/ram-lib-mapping.ts
```diff
@@ -0,0 +1,451 @@
...
+
+  if (bid) {
+    // Update associated token on Bid now that token has been minted
+    bid.token = generateContractSpecificId(
+      event.params.coreContract,
+      event.params.tokenId
+    );
+    bid.updatedAt = event.block.timestamp;
+    bid.save();
+  }
```
ryley-o, line 413: nit: in general, might be good to change this to a guard pattern where we log warning and return early if !bid (as a general pattern for handlers in this file)

tests/e2e/runner/supplemental_abis/MinterRAMV0.json
```diff
@@ -0,0 +1,1408 @@
+{
```
ryley-o, line 1: I think this is probably the only file that is needed in e2e supplemental abis?

tests/e2e/seed/scripts/deploy.ts
```diff
@@ -26,6 +26,7 @@ import { MinterSetPriceERC20V5__factory } from "../contracts/factories/MinterSet
 import { MinterSetPriceMerkleV5__factory } from "../contracts/factories/MinterSetPriceMerkleV5__factory";
 import { MinterSetPriceHolderV5__factory } from "../contracts/factories/MinterSetPriceHolderV5__factory";
 import { MinterSEAV1__factory } from "../contracts/factories/MinterSEAV1__factory";
+import { MinterRAMV0__factory } from "../contracts/factories/MinterRAMV0__factory";
```
ryley-o, line 29: you might want to pull the latest MinterRAMV0.json, because there have been some bug fixes in PR #1383

ryley-o (commented):
Overall looks good - didn't have time to give it a thorough review, but in general makes sense and is a clean implementation 🎊

abis-supplemental/MinterRAMV0.json
lindsgil: Thank you! Updated

schema.graphql
lindsgil: Good call, added this helper 🙏

src/ram-lib-mapping.ts
lindsgil: Good call! 🙏  Updated

src/ram-lib-mapping.ts
```diff
@@ -0,0 +1,451 @@
...
+  const ramBidId = generateRAMMinterBidId(
+    event.address.toHexString(),
+    fullProjectId,
+    event.params.bidId.toString()
+  );
+
+  // @dev: a bid with this ID should not already exist for the RAM Minter
+  const bid = new Bid(ramBidId);
+  // Create new account entity if one for the bidder doesn't exist
+  const bidderAccount = new Account(event.params.bidder.toHexString());
```
lyaunzbe, line 292: given the comment, should this be using a `load or create` style pattern?

lindsgil: keeping this consistent with the account logic in the account logic in this repo - I think we do this because accounts only have an ID and derived fields so if it does exist we are overwriting the existing ID with the same value, saving us from having to call load

src/ram-lib-mapping.ts
```diff
@@ -0,0 +1,532 @@
...
+  );
+  // @dev: a bid with this ID should not already exist for the RAM Minter
+  const bid = new Bid(ramBidId);
+  if (!bid) {
+    log.warning("Bid ID {} not found for minter {}", [
+      ramBidId,
+      event.address.toHexString()
+    ]);
+    return;
+  }
```
yoshiwarab, line 290: I'm a bit confused by this. `new Bid` will always create a new Bid entity so the `!bid` check below will never be true. Based on the comment I think this should be something like 
```suggestion
  let bid = Bid.load(ramBidId);
  if (bid) {
    log.warning("Bid ID {} already exists for minter {}", [
      ramBidId,
      event.address.toHexString()
    ]);
    return;
  }
  bid = new Bid(ramBidId);
```

src/ram-lib-mapping.ts
```diff
@@ -0,0 +1,532 @@
...
+  const bidderAccount = new Account(event.params.bidder.toHexString());
+  bidderAccount.save();
+
+  bid.project = minterProjectAndConfig.project.id;
+  bid.minter = event.address.toHexString();
+  bid.value = bidValue;
+  bid.bidder = bidderAccount.id;
+  bid.settled = false;
+  bid.slotIndex = event.params.slotIndex;
+  bid.winningBid = true;
```
yoshiwarab, line 301: Do we want winning bid to default to true? I feel like it should either align with the `settled` field (maybe a bit redundant) or made optional. 

Is there any case where winningBid and settled should have different values? I sort of imagine settled to be the equivalent of SEA winning bid but could be misunderstanding. WDYT?

tests/e2e/runner/__tests__/minter-suite-v2/ram-lib.test.ts
```diff
@@ -0,0 +1,750 @@
...
+      try {
+        await sharedMinterFilterContract
+          .connect(artist)
+          .removeMinterForProject(0, genArt721CoreAddress);
+      } catch (error) {
+        // swallow error in case of test failure
+      }
+    });
+
+    test("subgraph is updated after event emitted", async () => {
```
yoshiwarab, line 180: Nit: can you make the expected behavior a bit more specific here

tests/e2e/runner/__tests__/minter-suite-v2/ram-lib.test.ts
```diff
@@ -0,0 +1,750 @@
...
```
yoshiwarab, line 231: Nit: can you make the expected behavior a bit more specific here

tests/e2e/runner/__tests__/minter-suite-v2/ram-lib.test.ts
```diff
@@ -0,0 +1,750 @@
...
```
yoshiwarab, line 288: Nit: can you make the expected behavior a bit more specific here

tests/e2e/runner/__tests__/minter-suite-v2/ram-lib.test.ts
```diff
@@ -0,0 +1,750 @@
...
+      // validate project minter config in subgraph
+      await waitUntilSubgraphIsSynced(client);
+      // Get slot index for bid value
+      const slot10price = await minterRAMV0Contract.slotIndexToBidValue(
+        0, //_projectId
+        genArt721CoreAddress, // _coreContract
+        10 // _slotIndex
+      );
+      // Wait until auction starts (60s + 1s margin)
+      await new Promise((resolve) => {
```
yoshiwarab, line 319: I think it might be possible to use some hardhat network specific json rpc methods to jump ahead and avoid the wait here. Not totally sure how that will work with the graph node but probably worth a try. see https://hardhat.org/hardhat-network/docs/reference#evm_setnextblocktimestamp

yoshiwarab (changes requested):
Looks good with a couple minor changes, thank you! 

I have some reservations about using a single entity to represent both bid types, although I can't pinpoint a significant downside to managing the data downstream, particularly with only two bid types in play. The trade-off of a somewhat cluttered abstraction seems justified to avoid the complexity of introducing new tables and managing data synchronization separately. However, should we expand the variety of bids in the future, it would make sense to revisit our bid modeling strategy.

schema.graphql
```diff
@@ -479,7 +479,7 @@ type ProjectMinterConfiguration @entity {
 }
 
 type Bid @entity {
-  "Unique identifier made up of for SEA - {minter contract address}-{bidder}-{value}-{token} [TODO - do something different from RAM]"
+  "Unique identifier made up of for SEA - {minter contract address}-{bidder}-{value}-{token} and for RAM {minter contract address}-{full project ID}-{bidID}"
```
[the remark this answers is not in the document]
lindsgil: Good call! Added a bid type enum + updated the comments here to clarify RAM vs. SEA 🙏

src/ram-lib-mapping.ts
lindsgil: Good call, this makes more sense

src/ram-lib-mapping.ts
yoshiwarab: Sweet sounds good to keep! Thinking a bit more, it might make sense to have it be not optional and default to false instead of null to make it marginally easier to work with (no need to cast to a boolean if a function explicitly requires one) but ultimately ver minor difference so don't feel too strong.

lindsgil: Awesome, that makes sense to me! Let's do that - updated the logic + tests 🙏

tests/e2e/runner/__tests__/minter-suite-v2/ram-lib.test.ts
lindsgil: Thanks for flagging! Seems to work using `evm_increaseTime` - I added a helper in our test helpers file.

src/ram-lib-mapping.ts
```diff
@@ -0,0 +1,520 @@
...
+    JSONValue
+  > = createTypedMapFromEntries([
+    {
+      key: "startTime",
+      value: toJSONValue(event.params.timestampStart)
+    },
+    { key: "auctionEndTime", value: toJSONValue(event.params.timestampEnd) },
+    {
+      key: "numTokensInAuction",
+      value: toJSONValue(event.params.numTokensInAuction.toString())
```
ryley-o, line 177: nit: I think this could just be a number, right? (max of 1 million)

src/ram-lib-mapping.ts
```diff
@@ -0,0 +1,520 @@
...
+
+  const projectMinterConfig = minterProjectAndConfig.projectMinterConfiguration;
+  // update num tokens in auction detail in project minter configuration extraMinterDetails json field
+  let auctionConfigUpdateDetails: TypedMap<
+    string,
+    JSONValue
+  > = createTypedMapFromEntries([
+    {
+      key: "numTokensInAuction",
+      value: toJSONValue(event.params.numTokensInAuction.toString())
```
ryley-o, line 218: (same comment about storing this as string vs. number)

src/helpers.ts
```diff
@@ -198,6 +206,37 @@ export function loadOrCreateReceipt(
...
 
+/**
+ * helper function that returns the value of a bid in a given slot index, in Wei.
+ * used by the RAM Minter.
+ * @param basePrice Base price (or reserve price) of the auction in Wei, string representation
+ * @param slotIndex Slot index to query
+ * @returns slotBidValue Value of a bid in the slot, in Wei
+ */
+export function slotIndexToBidValue(
+  basePrice: string,
```
ryley-o, line 217: nit: any reason to not just have this input type be a BigInt?

src/ram-lib-mapping.ts
```diff
@@ -0,0 +1,520 @@
...
+      ramBidId,
+      event.address.toHexString()
+    ]);
+    return;
+  }
+  // Update slotIndex to null, value to 0
+  bid.value = BigInt.fromI32(0);
+  bid.slotIndex = null;
+  bid.updatedAt = event.block.timestamp;
+  bid.save();
```
ryley-o, line 338: While I think we could delete the bid from the store, updating it in a way that indicates it is deleted and syncing that to db makes sense to me.
Instead of zeroing out value and nulling slotIndex, WDYT about having a field `isRemoved` on the entity or something?

src/ram-lib-mapping.ts
```diff
@@ -0,0 +1,520 @@
...
```
ryley-o, line 515: (similar strategy here as bid deletion)

ryley-o (commented):
LGTM!
Added a few minor nits, and one question about what we should do upon bid deletion/refund

src/ram-lib-mapping.ts
lindsgil: Good call! Updated

src/helpers.ts
lindsgil: Good catch - updated this to BigInt 🙏

src/ram-lib-mapping.ts
lindsgil: Agreed! In favor of being explicit here.. I added the isRemoved field to the Bid entity - do you think it makes sense to update the sea mapping logic to set `isRemoved` to true when a user has been outbid?

src/ram-lib-mapping.ts
lindsgil: 🙏

src/ram-lib-mapping.ts
ryley-o: Yeah I like this and good call I think that behavior makes sense when a user is outbid on SEA as well

lindsgil: 🙏 Updated!

tests/e2e/runner/__tests__/minter-suite-v2/ram-lib.test.ts
```diff
@@ -724,21 +725,24 @@ describe("RAMLib event handling", () => {
       const refundedBid2Res = await getBidDetails(client, refundedBid2);
       const refundedBid3Res = await getBidDetails(client, refundedBid3);
 
-      expect(refundedBid1Res.slotIndex).toBe(null);
+      expect(refundedBid1Res.slotIndex).toBe("5");
+      expect(refundedBid1Res.value).toBe(slot5price.toString());
+      expect(refundedBid1Res.isRemoved).toBe(true);
```
ryley-o, line 730: nit: recommend adding a check for isRemoved == false at some point above

src/sea-lib-mapping.ts
```diff
@@ -242,6 +242,7 @@ export function handleAuctionBid(event: AuctionBid): void {
     const previousWinningBid = Bid.load(previousHighestBidId);
     if (previousWinningBid) {
       previousWinningBid.winningBid = false;
+      previousWinningBid.isRemoved = true;
```
ryley-o, line 245: do we need to set isRemoved when originally creating bid?

src/ram-lib-mapping.ts
```diff
@@ -0,0 +1,519 @@
...
+
+  bid.project = minterProjectAndConfig.project.id;
+  bid.bidType = "RAM";
+  bid.winningBid = false;
+  bid.minter = event.address.toHexString();
+  bid.value = bidValue;
+  bid.bidder = bidderAccount.id;
+  bid.settled = false;
+  bid.slotIndex = event.params.slotIndex;
+  bid.timestamp = event.block.timestamp;
```
ryley-o, line 296: I think we should set removed to false when creating, right?

tests/e2e/runner/__tests__/minter-suite-v2/sea-lib.test.ts
```diff
@@ -431,6 +432,7 @@ describe("SEALib event handling", () => {
       // Validate that the previous winning bid has been updated
       const previousWinningBidRes = await getBidDetails(client, bidId);
       expect(previousWinningBidRes.winningBid).toBe(false);
+      expect(previousWinningBidRes.isRemoved).toBe(true);
```
ryley-o, line 435: similar comment about checking case when isRemoved is false

ryley-o (approved):
lgtm! couple smol notes about the isRemoved is false case 🙏

tests/e2e/runner/__tests__/minter-suite-v2/ram-lib.test.ts
lindsgil: 🙏 🙏

yoshiwarab (approved):
LGTM 🚀

PR #288, opened 2024-04-11
Add base uri to create project to capture default value

Jbern16:
## Description of the change

baseURI if never updated is never synced -- so it causes confusion if it is set correctly if viewing purely from the CD.
This updates the create project to instantiate the default base uri returned from the contract

### DIFF
--- a/src/mapping-v3-core.ts
+++ b/src/mapping-v3-core.ts
@@ -460,7 +460,8 @@ function createProject(
     projectDetails.reverted ||
     projectScriptDetails.reverted ||
     projectStateData.reverted ||
-    projectArtistAddress.reverted
+    projectArtistAddress.reverted ||
+    projectBaseURI.reverted
   ) {
     log.warning("Failed to get project details for new project: {}-{}", [
       contractAddress,
@@ -488,6 +489,7 @@ function createProject(
   let project = new Project(
     generateContractSpecificId(contract._address, projectId)
   );
+
 
   project.active = false;
   project.artist = artist.id;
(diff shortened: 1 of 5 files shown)


src/mapping-v3-core.ts
```diff
@@ -454,6 +454,7 @@ function createProject(
   const projectScriptDetails = contract.try_projectScriptDetails(projectId);
   const projectStateData = contract.try_projectStateData(projectId);
   const projectArtistAddress = contract.try_projectIdToArtistAddress(projectId);
+  const projectBaseURI = contract.try_projectURIInfo(projectId);
 
   if (
     projectDetails.reverted ||
```
ryley-o, line 460: let's protect against `projectBaseURI` reverted case here

ryley-o (approved):
lgtm - tests might need updates 🙏 
Please do not merge into main until we are able to ship to decentralized (currently in the middle of deploying to different envs). Can definitely deploy to dev from this branch though!

tests/subgraph/mapping-v3-core/mapping-v3-core.test.ts
```diff
@@ -1389,6 +1399,12 @@ describe(`${coreType}: handleProjectUpdated`, () => {
         "artistAddress",
         artistAddress.toHexString()
       );
```
lindsgil, line 1401: Awesome! I think we need to add one more test for the default base URI when a project is added (not updated) right?

lindsgil (approved):

PR #289, opened 2024-04-22
Index core v3.2 event enum codes, flatten secondary financial data model

ryley-o:
## Description of the change

Update event handling to properly index new v3.2 events that emit PlatformUpdated and ProjectUpdated events with enum codes instead of bytes32-encoded strings.

This also flattens the provider secondary financial model to align with the required granularity on the v3.2 core contract. Relevant contract-level fields are marked as deprecated, and will be removed at a later date when downstream systems have been migrated.

This update is backwards compatible such that legacy V3 events or the new v3.2 events are handled equally. core contract version is now included in the `contract` entity, and populated on V3 core contracts, to support proper logic paths based on core version.

## Tests

The existing V3 core contract tests are updated to be passing. New test logic is added for coverage of:

- [x] the new schema fields on pre-v3.2 core contract versions (note: values null on V0, V1, and V2 core contracts, no updates needed)
- [x] The branched logic and enum events on v3.2+ contracts 

>note: existing core contracts are not e2e tests, so consistent strategy was used for these updates to prevent major re-write.

## Follow-on

Indexing ProjectRoyaltySplitterUpdated event will be in follow-on PR (able to be decoupled from this work).

It is recommended to update our e2e tests to use v3.2 core contracts in a future PR.

---
  - https://app.asana.com/0/0/1207198786683622


src/mapping-v3-core.ts
```diff
@@ -251,6 +255,30 @@ export function handleTransfer(event: Transfer): void {
   transfer.save();
 }
 
+// v3.2 core contract Enum ProjectUpdatedFields, encoded as bytes32 Bytes
```
yoshiwarab, line 258: Nit: can you update this comment to include a permalink to the relevant lines in the contracts repo for easy reference

src/mapping-v3-core.ts
```diff
@@ -432,6 +590,51 @@ function handleProjectSecondaryMarketRoyaltyPercentageUpdated(
   }
 }
 
+/**
+ * @notice Handle ProjectExternalAssetDependenciesLocked event. The event was
```
yoshiwarab, line 594: I think the start of this comment is for a different handler?

src/mapping-v3-core.ts
```diff
@@ -432,6 +590,51 @@ function handleProjectSecondaryMarketRoyaltyPercentageUpdated(
...
+): void {
+  // @dev event added in v3.2, so we can safely assume the contract conforms to
+  // the ProjectFinance interface. Use try/catch for redundant safety.
+  const contractAsV3_ProjectFinance = IGenArt721CoreContractV3_ProjectFinance.bind(
+    contract._address
+  );
+  const projectFinance = contractAsV3_ProjectFinance.try_projectIdToFinancials(
+    project.projectId
+  );
+  if (!projectFinance.reverted) {
```
yoshiwarab, line 614: Minor style nit: prefer early return

src/mapping-v3-core.ts
```diff
@@ -508,10 +711,64 @@ function createProject(
...
+  const coreVersion = contractEntity.version;
+  // version should never be null on a V3 contract
+  if (!coreVersion) {
+    log.error(
+      "V3 Contract entity for contract {} does not have a version set and has caused a subgraph error.",
+      [contract._address.toHexString()]
+    );
+    return null;
+  }
+  const isPreV3_2 = getIsPreV3_2(coreVersion as string);
```
yoshiwarab, line 732: Does this work fine now? I feel like I remember always having to use `changetype` for casting in assemblyscript

yoshiwarab (approved):

src/mapping-v3-core.ts
ryley-o: I think still have to use changetype if changing from one type to another - this was required to let AS compiler know `coreVersion` wasn't null for some reason, even though there is already an early return immediately before this statement 🤷 honestly a little confusing to me, but without, 
```
FAILURE 1 compile error(s)
ERROR TS2322: Type '~lib/string/String | null' is not assignable to type '~lib/string/String'.

   const isPreV3_2 = getIsPreV3_2(coreVersion);
```

src/mapping-v3-core.ts
ryley-o: good call, updated! [a475369](https://github.com/ArtBlocks/artblocks-subgraph/pull/289/commits/a47536973be141709ff17ddbbd662d5e335ffa26)

src/mapping-v3-core.ts
ryley-o: updated in [a475369](https://github.com/ArtBlocks/artblocks-subgraph/pull/289/commits/a47536973be141709ff17ddbbd662d5e335ffa26)

src/mapping-v3-core.ts
ryley-o: good call - I wish enum values were included in the ABIs for reference, but next best thing is this - updated in [a475369](https://github.com/ArtBlocks/artblocks-subgraph/pull/289/commits/a47536973be141709ff17ddbbd662d5e335ffa26)

Jbern16 (approved):

PR #335, opened 2025-03-13
[WIP]: PMPV0 Updates

lindsgil:
## Description of the change

Indexing `ProjectConfigured` and `TokenParamsConfigured` events to support PMP


src/mapping-pmp-v0.ts
```diff
@@ -0,0 +1,244 @@
...
+      } else if (previousPMPConfigCount.gt(BigInt.fromI32(0))) {
+        // delete all old pmp configs
+        for (let i = 0; i < previousPMPConfigCount.toI32(); i++) {
+          const prevPMPConfigId = generatePMPConfigId(
+            event.address,
+            event.params.projectId,
+            storedKeys[i]
+          );
+          store.remove("PMPConfig", prevPMPConfigId);
+        }
```
ryley-o, line 144: I think it's fine to only store the most recent PMP configs, because that is all that is ever used.
I was thinking if we should add `projectConfigNonce` to the `PMPProjectConfig` entity, then just store all PMP configs, relying on a `where` clause in graphql query to only get the latest PMP configs, but cleaning up and storing only most recent PMP configs seems most useful. I think that means there is an argument for not indexing `highestConfigNonce` on PMPConfig though...

ryley-o (commented):
🔥 🔥 🔥 this approach lgtm - I only found minor nits/tweaks/naming suggestions

src/mapping-pmp-v0.ts
lindsgil: Agreed - I couldn't think of a use case for storing project config revisions.. Removing highestConfigNonce from PMPConfig makes sense, also reduces complexity a bit - I removed that from the entity and updated `updatedAt` to `createdAt` on the PMPConfig entity 🙏

schema.graphql
```diff
@@ -929,3 +929,149 @@ type ProjectExternalAssetDependency @entity {
...
+  "String"
+  String
+}
+
+type PMPProjectConfig @entity {
+  "Unique identifier made up of {PMP_Plugin_Address}-{ProjectId}"
+  id: ID!
+
+  "Project associated with this PMP project config"
+  project: Project!
```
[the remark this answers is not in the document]
lindsgil: good call! added

src/mapping-pmp-v0.ts
```diff
@@ -0,0 +1,244 @@
...
+    const projectConfigId = generateProjectConfigId(
+      event.address,
+      event.params.projectId
+    );
+
+    let projectConfig = PMPProjectConfig.load(projectConfigId);
+    if (!projectConfig) {
+      projectConfig = new PMPProjectConfig(projectConfigId);
+      projectConfig.project = project.id;
+      projectConfig.save();
```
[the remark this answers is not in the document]
lindsgil: 🙏

src/mapping-pmp-v0.ts
```diff
@@ -0,0 +1,244 @@
...
+export function handleTokenParamsConfigured(
+  event: TokenParamsConfigured
+): void {
+  const tokenId = generateContractSpecificId(
+    event.params.coreContract,
+    event.params.tokenId
+  );
+
+  let token = Token.load(tokenId);
+  if (token) {
```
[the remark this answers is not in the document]
lindsgil: 🙏  updated!

src/mapping-pmp-v0.ts
```diff
@@ -0,0 +1,244 @@
...
+}
+
+export function handleProjectConfigured(event: ProjectConfigured): void {
+  let projectId = generateContractSpecificId(
+    event.params.coreContract,
+    event.params.projectId
+  );
+  let project = Project.load(projectId);
+
+  if (project) {
```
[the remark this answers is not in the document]
lindsgil: 🙏  updated!

src/mapping-pmp-v0.ts
```diff
@@ -0,0 +1,244 @@
...
+  "Unconfigured",
+  "Select",
+  "Bool",
+  "Uint256Range",
+  "Int256Range",
+  "DecimalRange",
+  "HexColor",
+  "Timestamp",
+  "String"
+];
```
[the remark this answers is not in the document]
lindsgil: moved! 🙏

schema.graphql
```diff
@@ -929,3 +929,149 @@ type ProjectExternalAssetDependency @entity {
...
+  configuredValue: String
+
+  "Artist-configured string value, if configured"
+  artistConfiguredValueString: String
+
+  "Non-Artist configured string value, if configured"
+  nonArtistConfiguredValueString: String
+
+  "Timestamp of last update"
+  updatedAt: BigInt!
```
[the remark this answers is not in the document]
lindsgil: nice! updated 🙏

subgraph.template.yaml
```diff
@@ -2289,6 +2289,39 @@ dataSources:
           handler: handleIAdminACLV0SuperAdminTransferred
       file: ./src/mapping-v3-core.ts
   {{/adminACLV0Contracts}}
+  {{#pmpV0Contracts}}
+  - kind: ethereum/contract
+    name: 'PMPV0{{#address}}-{{address}}{{/address}}'
+    network: {{network}}
+    source:
+      address: "{{#address}}{{address}}{{/address}}{{^address}}0x0000000000000000000000000000000000000000{{/address}}"
+      abi: PMPV0
```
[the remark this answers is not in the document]
lindsgil: 🙏 good call, thanks!! updated

yoshiwarab (approved):

src/mapping-pmp-v0.ts
```diff
@@ -0,0 +1,228 @@
...
+    event.params.coreContract,
+    event.params.tokenId
+  );
+
+  let token = Token.load(tokenId);
+
+  if (!token) {
+    // @dev While PMP is open to any NFT, we only index params on indexed projects
+    return;
+  }
```
ryley-o, line 30: I think instead of existence check on token, we should do existence check on project.
Technically PMPs can be defined pre-mint, so I think it would be okay create a PMP entity with a token that doesn't yet exist. as far as I can tell, nothing should break in our schema if that is the case...

schema.graphql
```diff
@@ -929,3 +932,146 @@ type ProjectExternalAssetDependency @entity {
...
+
+type PMP @entity(immutable: true) {
+  "Unique identifier made up of {PMP_Plugin_Address}-{tokenId}-{Key}-{tokenPMPNonce}"
+  id: ID!
+
+  "The key for this PMP"
+  key: String!
+
+  "The token associated with this PMP"
+  token: Token!
```
ryley-o, line 1051: @lindsgil should we at an @derivedFrom on the token entity to establish the array relationship?

src/mapping-pmp-v0.ts
lindsgil: ah! good catch, added the check for project 🙏

schema.graphql
lindsgil: 🙏 🙏  added!

src/mapping-pmp-v0.ts
```diff
@@ -0,0 +1,250 @@
...
+  for (let i = 0; i < event.params.pmpInputConfigs.length; i++) {
+    const config = event.params.pmpInputConfigs[i];
+    const pmpConfigId = generatePMPConfigId(
+      event.address,
+      event.params.coreContract,
+      event.params.projectId,
+      config.key
+    );
+
+    configKeys.push(pmpConfigId);
```
ryley-o, line 163: For consistency with the on-chain structs, I was expecting this would just store the PMP key value, and we would build the `prevPMPConfigId` when deleting, but this works just the same. if we do it this way, maybe it should be called `pmpConfigIds` in the schema?

tests/e2e/runner/__tests__/pmp/pmp.test.ts
```diff
@@ -0,0 +1,434 @@
...
+    expect(projectPmpConfigNextRes?.pmpConfigKeys).not.toContain(
+      oldPmpConfigKey
+    );
+
+    // validate project field
+    expect(projectNextRes?.pmpProjectConfig?.id).toBe(targetId);
+  });
+
+  describe("TokenParamsConfigured", () => {
+    test("creates a new PMP config for a token when configured", async () => {
```
ryley-o, line 290: it might be worth checking another case for when a PMP config is created for a non-existent token?

ryley-o (approved):
woohoo!
minor nits - one edge case around PMPs defined before a token exists that we can discuss more if you want!

src/mapping-pmp-v0.ts
```diff
@@ -0,0 +1,250 @@
...
+      pmpInput.key,
+      latestNonce
+    );
+
+    let pmp = new PMP(pmpId);
+    pmp.key = pmpInput.key;
+    if (token) {
+      // @dev it is valid for PMP to be created before token exists
+      pmp.token = token.id;
+    }
```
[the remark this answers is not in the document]
lindsgil: Ah, cool - that makes sense to me, appreciate the context here 🙏  I added a tokenId field to the PMP entity and updated the mapping + tests accordingly

src/mapping-pmp-v0.ts
lindsgil: ah makes sense! I had it that way at first and then updated it to the full id to reduce calls to the id builder helper function but prefer consistency with on-chain structs so reverting that back to just the key value!

tests/e2e/runner/__tests__/pmp/pmp.test.ts
lindsgil: Added!

src/mapping-pmp-v0.ts
```diff
@@ -0,0 +1,250 @@
...
+    );
+
+    pmp.tokenPMPNonce = latestNonce;
+    pmp.configuredParamType = currentConfiguredParamType;
+    pmp.configuredValue = pmpInput.configuredValue.toHexString();
+
+    if (pmpInput.configuringArtistString) {
+      pmp.artistConfiguredValueString = pmpInput.configuredValueString;
+    } else {
+      pmp.nonArtistConfiguredValueString = pmpInput.configuredValueString;
```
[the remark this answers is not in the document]
lindsgil: thanks! added 🙏

PR #336, opened 2025-03-24
graph pruning

ryley-o:
## Description of the change

Specify pruning of 500 blocks in subgraph schema to ensure low burden on indexers.

500 blocks is the upcoming limit supported by the upgrade indexer, and is all that should be needed by our typical indexing operations at Art Blocks.

## Complications

The pruning specification was not supported until manifest specVersion: 1.0.0, our subgraph has been defined on specVersion 0.2.0 since it was created. Quite a few updates are required to support this new specVersion, including updating dependencies, node versions, and the graph node e2e image.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18-alpine
+FROM node:20-alpine
 RUN apk add --no-cache curl
 RUN apk add --no-cache git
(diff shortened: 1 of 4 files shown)


ryley-o:
see latest commit CI e2e - graph-node has error to be fixed.

PR #340, opened 2025-04-08
Index pmp contract address in schema

lindsgil:
## Description of the change

Index pmp contract address in pmp project config schema


ryley-o (approved):
lgtm!

PR #353, opened 2026-02-25
New mint event

ryley-o:
## Description of the change

Update subgraph with new mint event handling.

The new mint event may be emitted from V3 core contracts, and includes token hash. this eliminates an rpc call in the subgraph handler, speeding up indexing of rapid mints.

Legacy testing is added, for consistency with our existing Mint event testing.

ref: https://github.com/ArtBlocks/artblocks-contracts/pull/1956


lindsgil (approved):
🔥 🔥

